PR #1, opened 2020-09-19
add workflow info in overview section

matt-maguire:
I added some content on workflows to the overview section:
- created a workflow overview section to illustrate "scene-referred" vs "display-referred"
- added content to scene-referred workflow section, largely based on Aurélien's "RGB Lab Which Module Help" article.
- added some basic remark to display-referred section telling users to look at the scene-referred section :-)
- fixed up the weights to order the pages

One problem is that the current theme has borderless tables, which makes them very hard to read. I created a shortcode to allow me to specify table-bordered table-striped properties from the bootstrap library. Maybe this is not the best way, we should probably fix the theme up at a global level.


paperdigits:
We should work quite hard to keep non-tabular data out of tables. This is why we've been using definition lists for all the module definitions. I've also though about not using shortcodes, because they're not valid markdown. I think I'd rather have the actual table HTML code in there, but lets see what @elstoc thinks.

Otherwise this looks great. Thanks for contributing!

matt-maguire:
ok, I see what you mean about definition lists. Let me try to align.

elstoc:
@matt-maguire this all looks great. I'll probably do some minor copy-editing once it's merged but apart from that it looks like a good addition. 

Please could you include some links though. Well I say "some" :-) . Certainly in the initial sections, wherever you mention a module or something that's explained more fully in another section, please can you include a link to that section. To keep the site as portable as possible please can you make sure that all the links use relative paths and all link directly to the `.md` file rather than the directory. I'll leave you to decide whether this is needed or not in your 'module order' section at the end.

TIA

matt-maguire:
I added in some links as requested, and also expanded the workflow overview and scene-referred sections slightly.

Not sure what to do with the "display-referred" aka legacy workflow section. We could maybe copy-paste a lot from the existing userguide, but it would overlap quite a bit with the scene-referred section, and maybe it just ends up confusing people. Any ideas what we should do here?

elstoc:
Thanks @matt-maguire that's great. If you are happy I'll merge this now. We can always tweak it again later if you like.  There will be a few rounds more copy-editing on the whole thing anyway before we're done.

matt-maguire:
Sure, let's merge it for now. Let me know if there are any other sections you want me to work on.

elstoc:
Thanks. I think most of the remaining ToDos are areas I don't use much. The larger ones that I'm currently avoiding doing are: tethering/map/slideshow/print views, tagging module and the remaining 'special topics'. So if you want to tackle any of those that would be much appreciated.

PR #2, opened 2020-09-25
add Slideshow section

matt-maguire:


elstoc:
Thanks @matt-maguire. Just two comments. Apart from the very top level, all headings and section names should be in lower case (overview and usage sections). Also, any keyboard shortcuts should be in `fixed width font`. Ideally blocks of keyboard shortcuts should be in a code block (see overview/user-interface/views for an example). I'm not precious about this if you think it works better as definition lists here (perhaps because there are multiple shortcuts that all do the same thing) but the shortcuts themselves should be fixed width.

elstoc:
Thanks @matt-maguire

PR #3, opened 2020-09-25
Created tone equalizer section

matt-maguire:
I created a section on tone equaliser, based more or less on the official guide, but with a but more of an overview description and diagram, and changed the language around a bit.


elstoc:
Working out how to write a reasonably concise explanation of the tone equalizer module that's clear and easy to understand is a massive thing to aim for and I commend you for tackling it. I don't think we're quite there yet though but I'm not sure how constructive I can be to work out how to get us there. So for the moment forgive me for commenting on some of the simpler stuff and avoiding, for now, how to fix the bigger issue of how to make this module more comprehensible to the uninitiated.

As we've discussed before, I'm not a fan of hand-made diagrams with text in them, just because we really want this stuff to be as translatable as we can get it without someone having to reproduce diagrams in another language. Also I'm not sure you've quite made it clear what the mask is doing and how it affects the operation of the module. Perhaps a worked example of an actual image might be a better way?

For defining module controls, the convention in this manual is to use definition lists rather than section headers. If a module is divided into sections/tabs we use separate section headers for each one and then define the controls within each section. Though because of the on-image controls we might need to break convention a bit here.

As for the 'official guide' I would consider it more "something someone wrote and got merged" rather than anything official or authoritative. I'm not convinced by the conversational tone (it sounds in places a bit like the subtitles from a youtube video) but again, I'm not sure that fixing the tone is enough.

I'm not sure I've been any help here. Perhaps @aurelienpierre might have some thoughts?

matt-maguire:
So, I tried to address the comments by:
- aligning to the heading-per-tab and definition list structure
- cleaned up some of the overly-conversational description
- updated the diagram to remove any english-specific text, using icons or callout labels, and move the description to the text of the moarkdown doc
- put some detail in the example scene to show it gets "blurred over" by the mask.

I think a worked example could be useful, but maybe this should go in the tutorial section of the manual. Need to find a nice image that nicely demonstrates how a mask works. Putting such a worked example in the reference section would introduce a large number of images, which means the section won't be very concise, and the big picture could be lost in the details of the example images. I suggest we add a reference in the introductory text of the module description if we later add such a worked example.

elstoc:
Thanks for your work on this @matt-maguire -- definitely an improvement over the current user manual

PR #7, opened 2020-09-26
[Issue #6] update tone-equalizer cursor to english, neutral background

### LINKED ISSUE #6: tone-equalizer-cursor.png is not English
The file `module-reference/processing-modules/tone-equalizer/tone-equalizer-cursor.png` looks like it's from a translated (French?) version of darktable and needs to be re-done in an English version, preferably with a more neutral background colour.

matt-maguire:
Fixes #6


elstoc:
Looks a bit larger and more "fuzzy" than the previous image - like it's been upscaled. Or is it me?

matt-maguire:
it is a screenshot, and has been upscaled by a factor of 2. I tried using a QHD monitor so get a higher resolution, but it didn't scale the cursor as I expected.

I tried applying some sharpening. If it is still not acceptable, we may neded to hack the cod to make the cursor bigger during the screen capture, or else we could manually draw the cursor to get a sharp image.

elstoc:
> I tried using a QHD monitor so get a higher resolution, but it didn't scale the cursor as I expected.

Indeed it doesn't seem to respond to either of the DPI settings. Ok we'll merge this one then. Thanks.

PR #8, opened 2020-09-27
update filmic rgb to cover v4 tabs/functions

matt-maguire:
The filmic section was describing filmic v3, so I updated it to align with filmic v4, which reoranised the tabs, introduced the highlights recovery features, and introduced the additional screens. I included a screenshot of the new zone mapping screen in the introduction section, since this seems to resonate with a lot of users to help them understand what the module does.

The workflow section seems to overlap with the workflow description in the manual's overview section. Also, I'm not sure if we still need the section for "filmic 2.6 users", are there still users around who haven't migrated to filmic v3.0 yet?


elstoc:
> I'm not sure if we still need the section for "filmic 2.6 users", are there still users around who haven't migrated to filmic v3.0 yet

I don't know, I still see reports on github and pixls.us of people still on 2.x versions. Having said that I don't really mind removing it.

elstoc:
Generally looks fine. A few minor points I'll cover in my copy-edit. Don't forget we're targeting American English (for reasons that escape me) so you have to spell everything wrong :-).

Thanks!

PR #9, opened 2020-09-27
fill out the Map section

matt-maguire:
based on content of existing userguide (probably not a lot more to say in fact)


elstoc:
Thanks @matt-maguire. I'll merge now and cover in my copy-edit but just a general thing: the current terminology in the manual refers to 'lib' modules as 'panels' - I'm trying to replace that with the word 'modules' throughout

PR #13, opened 2020-09-28
further copy-edit of tagging module, and add an image

matt-maguire:
Second pass...


elstoc:
Thanks. That reads much better now. Please can you look at the bulleted lists in the attach/detach/delete tag sections at the bottom. I'm still not 100% sure what some of them mean.

matt-maguire:
hope it make a bit more sense now?

elstoc:
Yep makes more sense now. Thanks @matt-maguire

PR #14, opened 2020-09-28
expand display-referred workflow

matt-maguire:
put some more info in the display-referred workflow section, mainly borrowing from current user guide, and highlight some differences between the scene-referred and display-referred workflows.


elstoc:
Thanks!

PR #16, opened 2020-09-29
export-metadata-upload: create content

matt-maguire:
created high-level summary of export process in workflow section.

### DIFF
--- a/content/overview/workflow/export-metadata-upload.md
+++ b/content/overview/workflow/export-metadata-upload.md
@@ -1,7 +1,27 @@
 ---
-title: exporting and uploading images with metadata (ToDo)
+title: exporting and uploading images with metadata
 id: export-metadata-upload
 draft: false
 weight: 50
 author: "people"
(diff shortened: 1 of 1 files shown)


elstoc:
Thanks @matt-maguire. It probably needs a bit more information about adding metadata to images though, given the title.

matt-maguire:
The mechanics of selecting which metadata to export should be covered in the export-slected lighttable module. In this section, we are more providing guidance on what high level activities should happen. Maybe I can mention something about the types of metadata that can be exported into the jpeg

elstoc:
I was thinking more along the lines of adding metadata to images, with reference to the tagging and metadata editor modules. But yeah - very high level.

matt-maguire:
That would more properly belong in the import, rate, tag section. At time of export, you are not normally adding metadata, you are selecting which metadata to write out.

elstoc:
Ah yes indeed you're right.

PR #18, opened 2020-09-30
wavelets: initial content

matt-maguire:
I had a go at making a very introductory article on wavelets, with a somewhat contrived example to try to illustrate what they can do.


elstoc:
A good first stab. Some initial suggestions / comments:

- Probably should sit immediately after the curves section (just change the weight in the metadata)
- As this is in the "interacting with modules" section it probably needs some reference to (and screenshots of) the two methods of representing/manipulating wavelet scales in dt modules (the 'equalizer' controls and retouch's wavelet decomposition)
- Probably needs another section on contrast equalizer (to go alongside the 'smoothing skin' section)
- The paragraph at the end about difference mode really only works in the contrast equalizer when you push the contrast of each of the scales in turn significantly

I won't rush into merging this one as it probably requires some consideration/discussion to get it right.

matt-maguire:
I think this section has maybe 2 aims: to give people a conceptual understanding of wavelets and what they are used for in darktable, and to try to "factor out" common parts of the modules so we don't have to keep repeating things. For wavelets, a similar coarse<->fine spline is used for the contrast equalizer and denoise (raw and profiled) module. The lowlight vision has a similar spline interface, but it operates in luminosity domain rather than wavelet domain. For some reason Aurélien has implemented his own spline for modules like tone equalizer and negadoctor (not sure why he didn't reuse the existing spline widget, maybe because he wants to offer a "simple" slider interface alongside the spline)

So, to @elstoc comments:
- it may make sense to include a section on how to deal with the spline curve for contrast equalizer and denoise modules. However, the contrast equalizer is a bit more complicated, because you have 2 splines on the one screen, but this doesn't exist in the denoise modules. So, for contrast equalizer, there will need to be some module-specific explanation. So we do it here, or in the contrast equalizer section?

- The interface for retouch is basically taken from gimp, and it is the only place that interface is used. So, perhaps this should be covered instead in the retouch module reference section?

- The example I chose for "skin smoothing" is not a tutorial on how to use retouch module, it is an example to illustrate the concept of how with wavelets you can apply processing to layers in a wavelet without impacting other layers, and the clearest way I could think was to use a blur, since it is visually quite obvious (I debated whether to use blending or not with the direct blur, but decided to do it to make the example look at least a bit more realistic). I suppose we could do a second example using contrast equalizer to show how contrast can be increased at one scale without impacting other scales, but this would be more subtle and may not show up so clearly in the images. We'd need to find a suitable image that can illustrate the concept clearly. But, it is not meant to be a tutorial, it is just meant to illustrate the impact of using wavelets.

- I thought I'd just include that "difference blending mode" as a practical tip, which can apply in general when trying to assess the impact of a module. For retouch module, it is not really required, but for other modules like contrast equalizer it can be useful. I suggest to move this tip to the contrast equalizer example, if we decide to include one. 

Maybe I've misunderstood the purpose of this section, in which case it makes sense to have a discussion about what should be covered where.

elstoc:
> However, the contrast equalizer is a bit more complicated, because you have 2 splines on the one screen, but this doesn't exist in the denoise modules

I meant something simpler like explaining that modules using these sort of splines allow you to modify the effect of a module based on the wavelet scale. Each node represents a different level of detail and you drag the nodes up/down to modify the effect (or something like that). So explaining basically what the splines are and how the user should interact with them.

> So, perhaps this should be covered instead in the retouch module reference section

In detail yes, but here, we should explain that retouch module allows us to choose the number of wavelet scales and operate on each one independently (with the detail of the actual operations covered in the module).

It's more a case of relating the general principles of wavelets to the common methods and controls that darktable uses to operate on them, covering the details of any commonalities so we're not repeating, for example, the "how to use splines" information.

> The example I chose for "skin smoothing" is not a tutorial on how to use retouch module, it is an example to illustrate the concept of how with wavelets you can apply processing to layers in a wavelet without impacting other layers [...]  I suppose we could do a second example using contrast equalizer to show how contrast can be increased at one scale without impacting other scales, but this would be more subtle and may not show up so clearly in the images.

It might then be better to remove this example to the retouch module and keep this section more general. Just cover the general principle that we can separate the image into detail levels and and that a module can operate on each of those details independently. Show your visual examples of those levels and relate them to the more general controls and leave detailed explanations to the separate modules.

> I suggest to move this tip to the contrast equalizer example, if we decide to include one.

Or maybe to the contrast equalizer section.

elstoc:
> For wavelets, a similar coarse<->fine spline is used for the contrast equalizer and denoise (raw and profiled) module. The lowlight vision has a similar spline interface, but it operates in luminosity domain rather than wavelet domain. For some reason Aurélien has implemented his own spline for modules like tone equalizer and negadoctor (not sure why he didn't reuse the existing spline widget, maybe because he wants to offer a "simple" slider interface alongside the spline)

Maybe we could have a separate 'splines' section to cover these commonalities without specific reference to wavelets.

matt-maguire:
I added a second example of using wavelets with cbntrast equalizer, showing how we can push the contrast of details in the foreground without impacting larger scale bokkeh in the background. Not sure how well it will show, contrast is a bit more subtle than blurring, especially if the resolution of the images gets reduced.

matt-maguire:
as requested, added in the .../wavelets directory cropped screenshots (not linked anywhere):
clean-retouch.png
clean-spline.png

elstoc:
Thanks @matt-maguire

PR #20, opened 2020-10-02
denoise-profiled: new content

matt-maguire:
Reworked existing user manual section and updated with the new Y0U0V0 mode. References the proposed wavelet section, which will be a broken link until PR #18 is merged.


elstoc:
On this and your other PR, please can you make sure your image links include the full (relative) path of the image. The way you've done it at the moment works, but only because of the way in which hugo renders its pages, and we want the markdown to be fully portable. So your links should be, for example ``./denoise-profiled/denoise-y0u0v0.png`` rather than just ``denoise-y0u0v0.png``

elstoc:
Thanks @matt-maguire. We're gradually getting towards the end of the ToDo list now. 

We're trying to avoid module screenshots where they're just a set of comboboxes and sliders so, unless/until we decide otherwise I'll remove the first screenshot as part of my copy-edit. Obviously the wavelet ones are more idiosynchratic so I'll leave those in.

PR #21, opened 2020-10-02
wavelets - elstoc version

elstoc:
My alternative edit to #18, which concentrates this section mainly on explaining the theory and the darktable controls used to apply module operations at the wavelet scale.


matt-maguire:
It is a lot more concise, but I think we've somehow lost the whole reason "why should I care about this? So you can make some trippy looking images, but what has it got to do my editing?"

matt-maguire:
looks good.

elstoc:
Thanks for your work on this @matt-maguire.

PR #23, opened 2020-10-03
[prefs] update to reflect status of recent darktable master

johnny-bit:
Couple updates in preferences window to match current master

Couple notes:
1. darktable seems to no longer support 32-bit systems, removed reference to bitness of systems.
2. map somewhere along the line lost the filtering of images, so the option was removed from manual
3. all changes reflect darktable _current master_ as of 03.10.2020 12:50 CEST


elstoc:
Looks good to me. Thanks @johnny-bit!

PR #26, opened 2020-10-04
color-zones: initial content

matt-maguire:
based mostly on the existing user guide, but updated the color picker and spline curve descriptions which seemed slightly out of date.


elstoc:
A couple of things, which are probably more comments on the current manual, but deserve addressing here.

- A clearer conceptual explanation is probably called for at the start. That first paragraph is the key and probably needs to be expanded a bit more. The examples are fine but a clearer explanation is needed first. I'd like to suggest some images might help but I'm not sure exactly how.

- This module probably should be referenced from the darkroom/interacting-with-modules/curves section (and maybe also the wavelets/splines section). The bits about the interpolation method can just include a link to the curves section, which contains a near-carbon-copy of those definitions.

matt-maguire:
I really struggled to find a way to explain the concept without ending up with ridiculously long and complex senetences. That's why I listed out some examples to try to illustrate it better. Maybe we need to treat this like it has 3 different modes, one for each x-axis, and explain each mode separately. It may make it more digestable.

I'll clean up the interpolation section -- I hadn't checked what you'd written in the "curves" section.

elstoc:
> I really struggled to find a way to explain the concept without ending up with ridiculously long and complex sentences.

I know what you mean. One of the reasons I put off working on it for so long. It's easy for those modules where the current manual is nearly there but for some of these more complicated modules you almost have to start from scratch to treat them in a non-confusing way. 

> Maybe we need to treat this like it has 3 different modes, one for each x-axis, and explain each mode separately. It may make it more digestable.

Yeah that might work. Feels like this one might take a few iterations and a bit of thought. Perhaps go through each of the tabs in detail using the default "select by hue" mode and then discuss the other "select by" modes more briefly, but with a set of side-by-side module screenshots for each, comparing them with the main discussion.

elstoc:
Yes that's looking much better. A few more comments...

Since you already have some screenshots of the pixel manipulation curves, it seems like we could sneak in some examples with actual curves, perhaps even matching the examples you have in your first section. So, darkening a blue sky, or changing red to yellow.

It's somewhat hard to see the nodes in the final screenshot - they might be easier to see around the blue/green area or on a different tab?

It might also be worth having a separate header for the color picker discussion

matt-maguire:
updated the curves wiuth some adjustment examples. II don't actually own a red porche, so I changed the colour of a pink toy instead :-)  In fact, these examples are a bit contrived. The hue-hue mapping is often used in video work with a vector scope to make slight adjustments to skin tones etc, but you wouldn't normally use it to completely change the color of an object in a photo. I dont think darktable has a vector scope widget, maybe that is a another feature enhancement.

elstoc:
ok let's go with that and I'll do a quick copy-edit. Thanks @matt-maguire.

PR #27, opened 2020-10-04
clipping: initial content

matt-maguire:
based on info in darktable PR #6253


elstoc:
I think this needs some definitions up-front -- mainly what is meant by the "allowed range" and "illegal values". 

I'm still not entirely clear exactly what the display is showing or what the module is helping me to do.

matt-maguire:
so RGB and luminance bounds are easiy to explain, but the gamut calculation is a bit more complicated. It is basically using a formula 
saturation = sqrt((x - L)^2 / (x^2 + L^2))
where x is one of the RGB channels, and L is the luminance (bascially a weighted average of the 3 RGB channels). If the normalised saturation exceeds the upper threshold (>98%) then it triggers the gamut warning.

I'm not sure where this formlua comes from, much of the literature seems to be behind a pay-wall. Might need to ask @aurelienpierre where he got this formula comes from.

matt-maguire:
actually, I thought about this a bit more, and I think I can see and I can see where this saturation formula is coming from now. Let me see if I can do a better write-up.

matt-maguire:
I tried to make the definitions a bit clearer. If this is still not clear enough, we'll probably need to go into some 2-dimensional and 3-dimensional CIE chromaticity charts to explain what a gamut is.

elstoc:
Much improved. Thanks @matt-maguire.

aurelienpierre:
> so RGB and luminance bounds are easiy to explain, but the gamut calculation is a bit more complicated. It is basically using a formula
> saturation = sqrt((x - L)^2 / (x^2 + L^2))
> where x is one of the RGB channels, and L is the luminance (bascially a weighted average of the 3 RGB channels). If the normalised saturation exceeds the upper threshold (>98%) then it triggers the gamut warning.
> 
> I'm not sure where this formlua comes from, much of the literature seems to be behind a pay-wall. Might need to ask @aurelienpierre where this formula comes from.

Saturation is computed as the normalized (by euclidean norm) radial distance between each RGB component and the corresponding luminance. The gist is, for a given pixel, the RGB channels need to be in a sphere centered at luminance of a certain radius to be in the gamut. That formula is empirical, though, but the output matches LittleCMS gamut alert good enough to be used as a scope. I wouldn't use it for actual pixel processing.

matt-maguire:
When I started drawing some diagrams (the x^2 + y^2 was a hint that I was looking at circles), it quickly became evident that this was checking whether the components lay within a sphere of radius L, and this sort of makes intuitive sense, since if we go beyond that, it means at least one of the channels must have been driven negative to arrive at that L value. Anyway, I just wanted to check my understanding to make sure I didn't write anything too silly. The user manual will need a technical review at some point anyway, to correct any misleading statements.

PR #28, opened 2020-10-04
contrast-equalizer: initial content

matt-maguire:
using the example fro mthe original user guide, just providing some clarifying comments here and there. The section seems to cover the key points about this setion, which, if read in conjunction with the wavelets overview, should give the user a basic enough understanding to get started with this module.


elstoc:
Trying to review but distracted by the funny picture! I think perhaps this could do with some sub-sections to break it up a little - it's a bit hard to navigate at the moment. Perhaps a sub-section on the spline controls and then one each for contrast adjustment, denoising, edge adjustment (which I'm still not sure I understand) and mix.

matt-maguire:
That's probably not a bad idea about the sub-sections. The example is from the official user guide, but if we are going to break into subsections, maybe it is an opportunity to add some more images to explain things better (in which case i'll need to find a raw I can use for that, since I don't have the original of that elephant).

elstoc:
Yeah I think this is one module where more words would be better, since it's quite powerful and the current manual doesn't exactly explain it very clearly.

matt-maguire:
I broke the document into separate sections per tab, and tried to show some visual examples. I'm having a lot of trouble finding a convincing example abomg my images to demonstrate the edge tab, maybe someone more experienced than I can provide a better example.

elstoc:
I agree the edges tab is hard to illustrate. I think, having played with it a bit, the best way to do it is to copy the existing manual and use very extreme settings. It would also be nice to find a way to do consistent screenshots that include the module and the key part of the image, without including the whole screen. So something like the following (not great as I've just thrown it together), where I've really overdone the luma contrast. 

The first image has the edge curve all the way down to zero:

Screenshot_2020-10-11_11-20-02

In the second it's at its default setting

Screenshot_2020-10-11_11-20-29

matt-maguire:
so I wasn't able to come up with a convincing example using the squirrel photo. I found an example in the research paper on which the contrast equalizer was based, and have cited the paper and included the example image of the edge awareness. I really don't think there is anything more I can do on this.

elstoc:
Ok that looks fine. Let's merge! Thanks

PR #35, opened 2020-10-08
color-zones: redo images with larger font

### LINKED ISSUE #34: color-zones: inconsistent images with small fonts
Problem:
In the color zones module documentation the first images are inconsistent with the last one. Also, the font size of the first images looks too small because they were taken from a wide panel (to increase their resolution). 

Resolution:
Either reduce the panel size or increase the font size and re-do the images.

matt-maguire:
Fixes #34  

Also regenerated all images for consistency.


elstoc:
Thanks @matt-maguire

PR #36, opened 2020-10-08
fix broken links

matt-maguire:
when building PDF, got some errors due to typos in links, so this allows PDF to build cleanly.


elstoc:
Thanks

PR #37, opened 2020-10-08
zoomable-lighttable: pad out content with example use case

matt-maguire:
flesh out the zoomable lighttable description with a sample use case and a hint about the darktable-generate-cache command.


elstoc:
Thanks @matt-maguire

PR #40, opened 2020-10-09
filmic-rgb: clean up wording on "exposing to the right"

### LINKED ISSUE #38: filmic RGB needs clarification around ETTR
Received some feedback from @paka on IRC that the wording around ETTR is a bit confusing.
- it is not mandatory to shoot ETTR, but it helps to optimise usage of sensor's DR (eg. for fast action may not have time to play with exposure settings)
- specular highlights can be clipped, this is a choice/tradeoff that has to be decided
- there was mention about the preview being dark, which was a confusing comment, because dark preview would suggest underexposure, whereas ETTR is the opposite. need to clarify what is meant here (tradeoff of highlights vs shadows with wide DR)
- ETTR refers to histogram in camera, not in darktable itself

If its causing confusion, needs to be cleaned up.

matt-maguire:
Fixes #38


elstoc:
Thanks @matt-maguire. Perhaps at some point later we could consider a separate section on capturing - most of these are good general recommendations rather than filmic-specific. For now, though, this is probably the best place for it.

matt-maguire:
agreed, maybe move it to the guides and tutorial section when that eventuates.

elstoc:
Yes, or even in the workflow section, perhaps with reference to external articles to keep the manual tidy.

PR #41, opened 2020-10-09
focus-peaking: initial content

matt-maguire:


elstoc:
Thanks @matt-maguire. Please could you re-do the image without the rest of the UI included - zoom out with Ctrl+scroll so that you get complete grey borders around the image and take a shot with the image just framed by those borders.

Avoiding including the UI wherever possible means that the images will last a bit longer if the UI changes (which it almost inevitably will).

matt-maguire:
I want to keep at least the bottom bit, because it shows the focus peaking icon is selected, and the user can see thwere it lives. I can crop the rest. Is that ok for you?

elstoc:
I dunno. You've already got an image of the icon. I think it would be enough to just add the words "on the bottom panel" to the text and I would still prefer just an image illustrating the feature.

matt-maguire:
ok. I took the same approach for color assessment, so I'll fix that one as well.

elstoc:
Thanks

PR #43, opened 2020-10-10
negadoctor: initial content

matt-maguire:


elstoc:
Thanks. I'll just do a very basic copy-edit on this one because I don't really know or care how it works :)

PR #44, opened 2020-10-10
search-and-group: initial content

matt-maguire:


elstoc:
Thanks @matt-maguire. Oddly enough, now that you've got used to American spellings, for some reason "favourites" is spelled correctly in darktable so I'll have to correct you the other way in my copy-edit!

elstoc:
Thinking about it some more I might move most of this content into the "module groups" utility module reference and keep this more of a "usage when editing images" section, then use the "module groups" section to more fully document AlicVB's recently-merged work.

matt-maguire:
yeah, I wasn't too sure how to split. I thought the module groups section could focus on the customitsation features. So in this section I just decided to describe what are default groups and what is the logic in mapping the modules to the groups.

PR #47, opened 2020-10-11
tone curves: fix color spaces list consistency

matt-maguire:
"Tone curves" section of official user manual was updated in PR 6323, but they botched it. There are in fact only 3 color spaces, but 4 items on the menu. Fixed our new user manual to make this clearer.


elstoc:
Thanks

PR #48, opened 2020-10-12
Input color profile

matt-maguire:
The input color profile module reference was a bit messed up. Have fixed the names of the 3 parameters for this module, and put a bit of description up front to explain what the input color profile module is in fact doing [do we need to mention XYZ transition color space? we could leave it out, but it is the standard paradigm for managing color space conversions, so may not be bad to mention it here].


matt-maguire:
These changes were actually prompted by the 3.4 To Do list, item #6499.

In the official manual, there is a section in "darkroom concepts" on color management, but we don't have a corresponding section in the new manual. Do we need to include one, or is the description in the input & output color profile modules sufficient? After all, this is a user manual for darktable, not a training course on photo development and color management.

elstoc:
I moved the color management section to the 'special topics' section because it seemed quite involved to put in the darkroom section.

content/special-topics/color-management/display-profile.md
```diff
@@ -6,7 +6,7 @@ draft: false
 author: "people"
 ---
 
- For darktable to accurately render colors on screen it needs to find the correct display profile for your monitor. In general this requires your monitor to be properly calibrated and profiled, and it needs the profile to be correctly installed on your system. darktable queries your X display server's xatom as well as the system service colord (if available) for the right profile. If required you can enforce a specific method in [`preferences` -> `miscellaneous`](../../preferences-settings/miscellaneous.md).
+ For darktable to accurately render colors on screen it needs to find the correct display profile for your monitor. In general this requires your monitor to be properly calibrated and profiled, and it needs the profile to be correctly installed on your system. darktable queries your X display server's xatom as well as the system service colord (if available) for the right profile. If required you can enforce a specific method in the [`preferences` -> `miscellaneous`](../../preferences-settings/miscellaneous.md) section.
```
elstoc, line 9: Not sure I agree with this one: preferences->miscellaneous is where you can enforce the method. The preferences -> miscellaneous section is where you can find out about it. This is the format I've used pretty much everywhere to refer to preferences sections.

matt-maguire: so by default, darktable will try to take into account both the X atom and the colord. However, if these aren't both set up properly, darktale can end up making a wrong decision. So, the purpose of this config item in the preferences is if you want to force it to excusively use one method or the other. Otherwise, just leave this setting at the default, and darktable will try to work it out using both moethds. i think you would not visit this setting unless you have a specific reason to override the default behaviour.

elstoc: As discussed it was just about the words before and after the link - another convention to take note of. I'll merge and put it back with the copy-edit

elstoc:
Thanks

PR #49, opened 2020-10-12
crop-and-rotate: update with the new tab containing margin controls

matt-maguire:
document new margins tab described in PR6233.


elstoc:
I removed "square" on purpose because I assumed probably everyone knows what a square is

matt-maguire:
squares are very important for instagram :-)

elstoc:
ok it doesn't bother me that much. We'll leave it in

PR #51, opened 2020-10-12
undo-redo: pr6055 add orientation to undoable items

matt-maguire:
only change required for pr6055 is to add "orientation" as something that can be undone.


elstoc:
Thanks, though it's probably not a complete list anyway

PR #53, opened 2020-10-13
pr6146: add performance mode & monochrome tagging to prefs

matt-maguire:
couple of missed items in prefs:
- auto monchrome tagging
- new performance mode setting


elstoc:
Thanks

PR #56, opened 2020-10-13
color-picker: align manual to new layout in pr6350

matt-maguire:
mention about new large swatch, reorganise sections to align better with gui layout.


elstoc:
Thanks

PR #59, opened 2020-10-13
Update HTML theme to make nav scroll independently and fix mobile layout

paperdigits:
Update HTML theme to make nav scroll independently and fix mobile layout


matt-maguire:
Very nice. Although, if I click on something in the TOC, the TOC keeps going back up to the top each time, which is a tad annoying.

paperdigits:
@matt-maguire not 100% finished with the theme yet, it is a WIP for sure.

PR #61, opened 2020-10-13
README: add info about images

matt-maguire:
Added some info about images. Not sure if we should describe the w25/w50/w75 hash tagging too?


elstoc:
> Not sure if we should describe the w25/w50/w75 hash tagging too?

We should indeed. The whole style/convention guide might get a bit long in the end - might be worth moving it to a separate file. Keep it here for now though

elstoc:
Thanks

PR #62, opened 2020-10-14
[Issue #57] document module resizing

### LINKED ISSUE #57: where to document adjustable height modules
PR6408 talks about being able to adjust the hiehgt of certain utility modules that contain lists of things by hovering the mouse over the list, then using `Ctrl-scroll` to change the maximum number of items to show in the list at one time.

This applies to modules on both lighttable and darkroom, and currently there is common area in the usermanual where to include such a description of module interaction. So, where to document it?

matt-maguire:
Fixed #57.

Add an overview section to the module reference and describe in there the new module resizing feature.

### DIFF
--- a/content/module-reference/overview.md
+++ b/content/module-reference/overview.md
@@ -1,0 +1,37 @@
+---
+title: overview
+id: module-reference-overview
+weight: 10
+draft: false
+author: "people"
+---
+
+The modules in this reference section have been broken down into two types:
+
+[processing modules](./processing-modules/_index.md)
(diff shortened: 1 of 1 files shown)


matt-maguire:
maybe double-check if it is allowed to reference an `_index.md` file like this. I see it has been done like this in other places, eg. the pixelpipe section.

elstoc:
Nothing wrong with referencing `_index.md` as far as I'm concerned. Thanks

PR #63, opened 2020-10-14
Full preview - initial content

matt-maguire:
The To-Do list indicated that the full preview section was missing from the current manual -- in fact, it now seems to be available, so I create some initial content for this feature.


elstoc:
Thanks. That's another ToDo done!

PR #71, opened 2020-10-16
history-stack: document pr#6520 safe history copy

matt-maguire:
Documents darktable-org/darktable#6520

### DIFF
--- a/content/module-reference/utility-modules/lighttable/history-stack.md
+++ b/content/module-reference/utility-modules/lighttable/history-stack.md
@@ -10,10 +10,10 @@ Manipulate the history stack of one or more selected images.
 
 # module controls
 
-copy
+copy parts...
 : Copy the history stack of the selected image. A dialog appears within which you may choose which items from the history stack to include. If more than one image is selected, the history stack is taken from the image that was selected first. Double-clicking on an item copies that item only and immediately closes the dialog.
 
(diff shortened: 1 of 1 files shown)


elstoc:
Let's wait and see how this feature is received before merging this.

junkyardsparkle:
I agree that there could be some backlash against this change, we'll see. This makes me ask: would it be within the scope of this project to maintain a prominent list of potential "gotchas" for each version to help keep people from encountering these kinds of changes as (possibly slow-to-be-noticed) "bugs"? I have a feeling people don't always read the release notes...

matt-maguire:
If they don't read the release notes, they are unlikely to read the udser manual. That said, some sort of a "FAQ" or "Hot Tips" in the guides and tutorials section could be a useful reference for people, which could capture some of those "gotchas".

elstoc:
I think that something like that should be stored separately, perhaps in a wiki on the main darktable project, with a link from the user manual.

elstoc:
Guess this needs updating now. Suspect this feature isn't likely to change again before 3.4.

matt-maguire:
Looks like the list of "safe" modules hasn't changed at all, only change seems to be the button names. Updated the history stack utility module page accordingly.

elstoc:
Thanks

PR #73, opened 2020-10-16
guides-tutorials: initial content

matt-maguire:
Some initial content identifying some key learning resources for darktable


elstoc:
Thanks @matt-maguire

PR #74, opened 2020-10-16
Add a few more changes from PR6499

junkyardsparkle:
Just a couple of little things probably worth pulling over. So, now that the color management section has been moved to "special topics", should it be more detailed and less of an "overview"?

### DIFF
--- a/content/special-topics/color-management/color-spaces.md
+++ b/content/special-topics/color-management/color-spaces.md
@@ -19,7 +19,8 @@ between [_demosaic_](../../module-reference/processing-modules/demosaic.md) and 
 : Image is in RGB format within the color space of the specific camera or input file.
 
 between [_input color profile_](../../module-reference/processing-modules/input-color-profile.md) and [_output color profile_](../../module-reference/processing-modules/output-color-profile.md) 	  	
-: Image is in Lab format. This is a very large universal color space which covers all colors visible to the human eye (and even more). As darktable processes images in 4x32-bit floating point buffers, it can handle the Lab color space without risking banding or tonal breaks.
+: Image is in the space defined by the selected working profile (linear Rec2020 RGB by default). As darktable processes images in 4x32-bit floating point buffers, we can handle large working color spaces without risking banding or tonal breaks.
+
 
 after [_output color profile_](../../module-reference/processing-modules/output-color-profile.md) 	  	
 : Image is in RGB format as defined by the selected display or output ICC profile.
--- a/content/special-topics/color-management/overview.md
+++ b/content/special-topics/color-management/overview.md
@@ -12,4 +12,4 @@ author: "people"
 
 - darktable automatically reads the display profile of your monitor (if properly configured) for accurate on-screen color rendition. Multi-screen setups are fully supported as long as a system service like colord is in place and properly set up to inform darktable of the correct monitor profile.
 
-- Output files can be encoded in one of darktable's built-in profiles, including sRGB, Adobe, ProPhoto, Rec. 2020 and Rec. 709 standard RGB color spaces, or into a color space supplied by the user as an ICC profile.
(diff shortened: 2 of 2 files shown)


content/special-topics/color-management/overview.md
```diff
@@ -12,4 +12,4 @@ author: "people"
 
 - darktable automatically reads the display profile of your monitor (if properly configured) for accurate on-screen color rendition. Multi-screen setups are fully supported as long as a system service like colord is in place and properly set up to inform darktable of the correct monitor profile.
 
-- Output files can be encoded in one of darktable's built-in profiles (e.g. sRGB or AdobeRGB) or into any other output color space that the supplied by the user as an ICC profile.
+- Output files can be encoded in one of darktable's built-in profiles, including sRGB, Adobe, ProPhoto, Rec. 2020 and Rec. 709 standard RGB color spaces, or into a color space supplied by the user as an ICC profile.
```
matt-maguire, line 15: There's no such space as "Adobe", it should be "AdobeRGB". This together with sRGB are the most common use cases for output profile, so I don't think it is worth trying to list some of the others unless it is going to be done comprehensively. We can probably assume that if someone wants to use one of those other color spaces, they are an advanced user, and can see in the drop-down list what their options are.

content/special-topics/color-management/color-spaces.md
```diff
@@ -19,7 +19,8 @@ between [_demosaic_](../../module-reference/processing-modules/demosaic.md) and
 : Image is in RGB format within the color space of the specific camera or input file.
 
 between [_input color profile_](../../module-reference/processing-modules/input-color-profile.md) and [_output color profile_](../../module-reference/processing-modules/output-color-profile.md) 	  	
-: Image is in Lab format. This is a very large universal color space which covers all colors visible to the human eye (and even more). As darktable processes images in 4x32-bit floating point buffers, it can handle the Lab color space without risking banding or tonal breaks.
+: Image is in the space defined by the selected working profile (linear Rec2020 RGB by default). As darktable processes images in 4x32-bit floating point buffers, we can handle large working color spaces without risking banding or tonal breaks.
```
matt-maguire, line 22: good catch!

content/special-topics/color-management/overview.md
junkyardsparkle: Fine with me, I only added them to the orginal PR at the request of Pascal Obry.

junkyardsparkle: But there's still that typo in the last sentence to fix. ;-)

elstoc:
Thanks @junkyardsparkle. Welcome to the party :)

matt-maguire:
> ...should it be more detailed and less of an "overview"?

This is a user manual for darktable, not a textbook on image processing, so we probably don't want to go into too much detail on color space theory. Some practical tips for common use cases could be useful though, eg. how to chose between sRGB and AdobeRGB, what color space should I use if I want to edit my image in GIMP, etc..

elstoc:
If we can find some useful external resources though, no harm in adding some links.

paperdigits:
Yes I agree with Matt. Every word we write here, we have to keep 
up-to-date, and other people have to work to translate.

We should strive to write as little as possible in order to get the 
point across. For common image processing concepts, we should try not to 
include that sort of stuff.

A while ago I started another repo to write all the camera and image 
processing stuff down so people could use it, but I never got that far...

junkyardsparkle:
> If we can find some useful external resources though, no harm in adding some links.

That was actually going to be one of my next questions... I didn't copy over the wikipedia etc links... where do those fall in the definition of "useful external resources"?

junkyardsparkle:
> This is a user manual for darktable, not a textbook on image processing, so we probably don't want to go into too much detail on color space theory.

Yep, agreed. By "more dtail" I was thinking more of references to specific aspects of the darktable user interface, preference settings, etc... but of course all of that falls under "stuff that has to be maintained", so... :man_shrugging:

PR #77, opened 2020-10-17
scene-referred-workflow: small edits, move non-recommended modules to…

matt-maguire:
… end

- Move the non-recommended modules to the end of the section
- clarify and lonk to "non-local means" module to use when no profiled support
- other minor edits.


elstoc:
Yes that's definitely the better order. Thanks.

PR #78, opened 2020-10-19
filmic-rgb: cut out unnecessary gumf at the end

matt-maguire:
- chop off workflow-related information at the end, and instead put a couple of the useful remarks from there into the usage section.
- be more consistent with italicising _filmic rgb_.


elstoc:
Looks good to me. Thanks

PR #81, opened 2020-10-19
exposure: offer alternatives to using black level slider

matt-maguire:
after conversation with @dtschfr in IRC, add a remark in exposure module how to deepen the blacks wirthout using black level slider.


elstoc:
Thanks

PR #91, opened 2020-10-20
Add intro splash page

### LINKED ISSUE #90: Write an introduction to the user manual
The user manual needs some introductory message/preamble explaining what darktable is for, which can be shown in the right hand panel when entering the user manual from the top level.

matt-maguire:
fixes #90 

Add a splash page to user manual based on branding message from website.

Not sure if we want to add some additional information, like "this is the user manual, applicable to darktable v3.4" or something like that...


elstoc:
Certainly a big improvement over the blank page. Let's keep it at that for the moment. Thanks @matt-maguire.

PR #102, opened 2020-10-26
filmic-rgb: update preserve chrominance settings

matt-maguire:
updated description of _preserve chrominance_ settings, in line with filmic FAQ


elstoc:
Looks good. Thanks

PR #103, opened 2020-10-27
[Issue #99] improve smoothing diameter and preserve details descriptions

### LINKED ISSUE #99: Tone EQ
module-reference/processing-modules/tone-equalizer/

> This controls how much of the surrounding image to take into account when calculating the mask’s blur at a particular point. By default it takes 25% of the surrounding into account.

'25% of the surrounding(s)' doesn't really mean anything. Checking the code, the blurring radius is defined in terms of image dimensions (`d->blending` is the value entered on the UI as 'smoothing diameter'), even though the exact calculation is not that important from a user's point of view:
```
  const int max_size = (piece->iwidth > piece->iheight) ? piece->iwidth : piece->iheight;
  const float diameter = d->blending * max_size * roi_in->scale;
  const int radius = (int)((diameter - 1.0f) / ( 2.0f));
```
So, it could perhaps be worded as 'This controls how much of the surrounding image to take into account when calculating the mask’s blur at a particular point, in relation to the image dimensions. The default value is 25% of the longest side.'
Maybe including the values Aurélien considered typical would make sense: 'Typical blurring radii are around 1-5% [for EIGF], while the original GF is better between 10-25%' (https://discuss.pixls.us/t/a-tone-equalizer-in-darktable/10678/10).

matt-maguire:
Fixes #99 

implement improvement suggestions and document new EIGF functionality.


elstoc:
Thanks

PR #110, opened 2020-10-30
improve color management sections

aurelienpierre:


content/special-topics/color-management/color-spaces.md
```diff
@@ -6,21 +6,27 @@ draft: false
...
 
 Each module has a position in the pixelpipe which tells you which color space the module lives in:
 
 up to [_demosaic_](../../module-reference/processing-modules/demosaic.md)
-: Image is in raw data format with only latent colors. Each pixel carries lightness and color information for only one base color. Bear in mind that some of the modules in this part of the pipe can also act on non-raw input images in RGB format with full information on all three color channels.
+: Image is in raw data format with only latent data. Each pixel carries a single intensity for one single primary color, and camera primaries are very different from human primaries. Bear in mind that some of the modules in this part of the pipe can also act on non-raw input images in RGB format with full information on all three color channels.
```
elstoc, line 22: I'm not sure that "Image is in raw data with only latent data" really says anything. Are you ok with removing this sentence entirely?

content/special-topics/color-management/color-spaces.md
```diff
@@ -6,21 +6,27 @@ draft: false
...
+darktable's input images are either RGB files (like JPEGs or TIFFs) or camera raws. Both store visual information as a combination of primary colours like red, green and blue which, together, describe a light emission to be later recreated by a display.
+
+It is important to mention that colour only exists in the human brain, as a product of the decomposition of a light emission by the cone cells in the retina. What RGB files do is describing light emissions with a similar method, for technical reasons. However, the definition of each primary colour depends on the properties of each RGB space and might look very different to human eyes from what humans call "red", "green" or "blue".
+
+Notice of warning, RGB is not really a color, but rather a light emission being encoded for computational purposes, that will later be physically created by your display, and interpreted as colour only once it enters your brain through your retina. Calling RGB "colour" is a shortcut.
```
elstoc, line 13: I'm not sure about the sentences "RGB is not really a color" and "Calling RGB "color" is a shortcut". Agree with the sentiment just not sure about the term RGB to represent a single "color". We're using RGB to refer to the color "space". Perhaps something like "A single RGB value is not really a color" but it feels a bit clumsy.

Aside: we're using American spellings to be consistent with the application, so "color" rather than "colour".

elstoc (commented):
A couple of minor queries. There are some spelling and grammar errors as well but I'll fix those in my copy-edit.

content/special-topics/color-management/color-spaces.md
aurelienpierre: What I mean is that the raw file needs some reconstruction work, it's not yet an image + the raw RGB is really not RGB in the way we expect it. The formulation is perhaps unclear, but I would like to have that stated somewhere because it's important.

elstoc: How about "The raw image information does not yet constitute an "image" but merely "data" about the light captured by the camera. Each pixel carries..."

content/special-topics/color-management/color-spaces.md
aurelienpierre: Well, an RGB vector can encode a colo~~u~~r, if you abide by the web definition of color. For example, RGB = { 0.75, 0.75, 0.5 } "is" orange and you will find many online converters that will do it for you. The original text mentioned "RGB colors". 

What I'm trying to fight here is the idea that any RGB vector is **directly** a color, because if people stick to that, many things in the scene-referred workflow will make zero sense. It's much easier to conceptualize RGB as lights you dim and mix to reconstruct the original spectrum. That's also something you can experiment with IRL with dimable light bulbs and gels. 

Point being, we see colors, but we work on light, and the light -> color conversion is very different depending on the sensor.

content/special-topics/color-management/color-spaces.md
matt-maguire: The raw image information does not yet represent a *viewable* image, but is merely consisting of data about the light captured by the camera. Each pixel carries...

elstoc: @matt-maguire that sounds good to me

elstoc:
@aurelienpierre: is this ready to review/merge now?

PR #112, opened 2020-11-01
add diagram to clarify about input vs output masking

matt-maguire:
prompted by the recent discussions around input vs output masking, I thought I'd add a diagram to dtdocs to try to make this mechanism a bit clearer.


elstoc:
I think we need to find a way to simplify that diagram, though I'm not yet sure how.

matt-maguire:
So, the whole orange box is mixing the in and out images together, and the alpha signal from the mask is like the balance control on your stereo, controlling the balance between original image and processed image. We can take the mask bit out of the orange box, and the rest of the stuff in the orange box can be abstracted away. The detail there is perhaps misleading awayway, sunce there are a variety of different blending modes.

elstoc:
I think you can just explain the alpha in the text or indicate it as a single input into the last stage.

elstoc:
Perhaps some shapes as a third input into the mask or will that overcomplicate the diagram?

matt-maguire:
ok, tired to make it look more like a balance control.

elstoc:
Forgive the crude sketch but I think moving the arrows so that they all originate from boxes and representing the input/output and masks as images might make it clearer:

PXL_20201101_114327958

Then you have paragraphs to talk about input, operation, output, masking, blending moving from left to right.

matt-maguire:
So I've tried to capture a more precise representation of the masking/blending process, and I updated the text. I've put in some high-level formulas of what is going on, but we need to decide whether we want to simplify or not.

elstoc:
It needs some more thought. I think it looks very complex, but then of course it *is* very complex, and I'm not sure how to make it simpler or clearer. I do still think that the P(I) and B(I, P(I)) notation, though accurate, may put people off. Perhaps subscripts might be better (can we do subscripts in markdown?). Also I don't think the tool icon next to the "g" really adds anything. But all that's just nibbling at the edges of the larger problem of how to make this [as] easy [as possible] to understand.

The other initial thought is that this now repeats quite a lot of information in the "anatomy of a module" section so I think it might be better incorporated there (removing explicit reference to the pixelpipe because it's not yet been covered).

For now I think the diagram has at least got it clearer in my head how it works so I can understand it enough to start considering the best way forward. Please could you attach the source for the image to a comment on the PR in case I want to have a play and try some things out.

matt-maguire:
I think we can avoid that notation, just replace it with regular labels. Subscripts require HTML to be added <sub></sub>, but I think people might be scared off by that too. Maybe just use a simple letter or number.

The mask currently has 2 inputs, for parameteric masks based on input and output, but I think we can add a third input which is a set of drawn shapes. We may not want to go into too much details about masks here (including, exclusive, etc.) sine there is a separate section for that, and this is just an overview.

The "g" I just wanted to indicate it is a user control -- probably a slider image is better, but I just used whatever clipart was handy in the drawing tool.

Glad it's got the wheels turning in your head, hopefully we can land on something useful to all.

elstoc:
Thanks Matt.

> I just used whatever clipart was handy in the drawing tool

Those images don't work in my version of draw so we might need to replace them with embedded image files.

matt-maguire:
I had a look at graphviz and inkscape, and I still think LibreOffice draw is the simplest tool going forward.

I rejigged the diagram to eliminate clipart, give it more of a darktable "theme", and simplify it a bit (eg. remove the formulas).

elstoc:
Replaced by #116

PR #114, opened 2020-11-01
Style shortcuts

ralfbrown:
Point out that styles can also be assigned to shortcut keys.

There should be mention of this in shortcuts.md, but that's a bigger insertion than the equivalent mention for presets was.  Probably a new subsection to explain that global -> styles gives a list of currently-defined styles for assignment to shortcut keys.


elstoc:
Thanks @ralfbrown. I'll fix it up in my copy-edit on this one, but please see the README for conventions for links - specifically, links to preferences tabs use ``[`preferences > shrotcuts`](../../preferences-settings/shortcuts.md)``

elstoc:
Actually, I think it probably is worth expanding shortcuts.md. I'll do that while I'm there.

PR #117, opened 2020-11-05
edit-overview: convert diagram from ascii-art to image

matt-maguire:
Before we had some ascii-based diagram for scene-referred vs display-referred pixelpipe. This is quite difficult to handle in the translation tools, so I converted it to a graphical representation that also make the diagram stand out better.


elstoc:
Thanks

PR #118, opened 2020-11-06
tone-equalizer: fix image links and make diag horizontal

matt-maguire:
try to align tone equalizer diagram to the horizontal stripped format we used for other diagrams. Also fixed image links so they work better with github.


elstoc:
Yes that's definitely an improvement - the whole 1,2,3,4,A,B,C combination always confused me a little. I think I would prefer the section splits a little different though...

Screenshot_2020-11-06_13-26-26

[tone-equalizer-overview.zip](https://github.com/elstoc/dtdocs/files/5501448/tone-equalizer-overview.zip)

1. Generate the mask
2. Generate the curve
3. Combine the mask and the curve

matt-maguire:
Yeah, that's much better. I'll update it and fix the words.

elstoc:
Thanks

PR #122, opened 2020-11-11
color-calibration: explain more about valid slider settings on grey tab

matt-maguire:
In darktable-org/darktable#6854 it was explained that the sliders on the grey tab of *color-calibration* need to add up to something non-zero, so this updates the doc to clarify valid slider settings.

### DIFF
--- a/content/module-reference/processing-modules/color-calibration.md
+++ b/content/module-reference/processing-modules/color-calibration.md
@@ -150,7 +150,7 @@ normalize channels
 This tab is useful to convert a color image into a greyscale image. 
 
 input red/green/blue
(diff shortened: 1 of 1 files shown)


elstoc:
That's only true if you don't choose to normalize channels though

matt-maguire:
Really? When I tested, it seemed the monochrome treatment was disabled even when normalisation was off. It is a bit of a pathological case though, because without normalisation as the sum approaches zero the image will be quite dim.

elstoc:
Sorry I meant that the bit you've put about the image going black with negative values isn't correct with normalisation.

matt-maguire:
Yes, you're right, because dividing a negative with a negative gives you a positive. So, I just removed the comment about negative totals altogether, in order to not overcomplicate things.

elstoc:
ok thanks

PR #124, opened 2020-11-15
color-calibration: add some more explanations about conflict warnings…

matt-maguire:
… and bypassing CAT

Prompted by the discussion in darktable-org/darktable#6921


elstoc:
Thanks. I'll get round to a copy-edit soon :)

chhil:
Would there be a case when you don't want to bypass it in the seond instance?

If ther is no case for it, may be creating a new instance should set the parameter appropriately to bypass?

matt-maguire:
yes, you may want multiple instances doing different white balancing treatments in different parts of the image, using masks to guide which part of the image each instance operates on. The masks could be drawn parametric, gradient, etc.. It's really up to the user to specify what they want from the module, expecially in such advanced use cases.

PR #142, opened 2020-11-20
suggestions and corrections for overview section

quovadit:
another image-version for screen-layout including the border-only png, some text-corrections (e.g. typos) and some text-suggestions


elstoc:
@quovadit: Please can you keep your PRs separate. I think we're still considering the best way to do the screen layout but most of the corrections in your second commit look fine to merge. It would be much easier to do this if they remained independent PRs.

content/module-reference/utility-modules/shared/filmstrip.md
```diff
@@ -12,9 +12,9 @@ The filmstrip can be used to quickly switch between images in any darktable view
 
 The filmstrip can be switched on and off using the shortcut `Ctrl+F`. The height of the filmstrip panel can be changed by clicking and dragging its top border.
 
-Quickly navigate through the images in the filmstrip by scrolling with the mouse. In the darkroom you can change the photo currently being processed by double clicking on another image in the filmstrip.
+Quickly navigate through the images in the filmstrip by scrolling with the mouse. In the darkroom you can change the photo currently being processed by clicking on another image in the filmstrip.
 
-In the darkroom, the image currently being processed is selected and highlighted. Single click to select any other image from the filmstrip without changing the image being processed.
+In the darkroom, the image currently being processed is selected and highlighted. Use Mouse over to select any other image from the filmstrip without changing the image being processed.
```
elstoc, line 17: I think either works. Perhaps we should say "hover over or single-click on an image..."

content/overview/user-interface/top-panel.md
```diff
@@ -8,26 +8,35 @@ author: "people"
...
 : Define what information is displayed over thumbnails in the lighttable/filmstrip.
+: You can define different settings depending on the thumbnail size (set sizes categories in [`preferences > lighttable`](../../preferences-settings/lighttable.md))
 
 context-sensitive help
+![top panel - help](./top-panel/top-panel_help.png)
 : Click on this icon and then click on a control element to be directed to the appropriate online help page.
 
 preferences
+![top panel - preferences](./top-panel/top-panel_preferences.png)
 : Open the [preferences & settings](../../preferences-settings/_index.md) dialog
```
elstoc, line 42: I'm not convinced about this. I preferred what was there before. Perhaps if we had two screenshots - one for the left-hand side and one for the right-hand-side.

elstoc (changes requested):
Mostly looks fine - just a couple of comments. Please can you squash to a single commit so that we lose the screen layout change from the history entirely.

content/module-reference/utility-modules/shared/filmstrip.md
quovadit: Single click now changes the image currently being processed. So what was single click before is mouse over now (only select for ratings etc but do not set it as image to be processed)

elstoc: hmm there's something fishy about that. I'm not sure that's intentional.

content/overview/user-interface/top-panel.md
quovadit: for me it's difficult to see, which icon refers to a function:
- the star is not clear for me to define the thumbnail overlays
- I did not notice the possibility to change the sort order with the second arrow for a long time

So in my opinion it is important to split the icons (not only left and right)

elstoc: Maybe most of the problem I have with this is that all of your screenshots are different heights/widths so it looks a bit ugly. Try tidying up the screenshots so they're consistent and we'll see what it looks like.

content/module-reference/utility-modules/shared/filmstrip.md
elstoc: I've raised an issue https://github.com/darktable-org/darktable/issues/7000

elstoc: Ok so it's `Alt`+click to select an image without changing image, and then `Ctrl`+click and `Shift`+click to extend the selection. Not sure why we have to make it work that way but it seems it's intentional.

content/overview/user-interface/top-panel.md
elstoc: > I did not notice the possibility to change the sort order with the second arrow for a long time

Perhaps keep it as it was and just add this icon and a third (ascending/descending) definition (I agree it's not clear)

quovadit: I agree that the images weren't beautiful, I've now cleaned up the icons and removed the drop-downs, for me that looks fine and is informative as well

What do you think?

quovadit: maybe make the background of the icons transparent?
(and of course #inline is not necessary anymore, because I've chosen icon finally)

quovadit: no, transparent is even worse...
but it's hard to see the icons with the dark background. 
What about a lighter one, even if it is not the default-dt-style?

elstoc: Yes that's better but please can you put the previous version of the top panel image back as it had less dead space in the middle

content/module-reference/utility-modules/shared/filmstrip.md
elstoc: Let's correct this for now to talk about the existing behaviour and we'll change it back again if it gets changed back as a result of the issue.

content/overview/user-interface/top-panel.md
quovadit: dead space in top panel is removed
I now put transparent icons to the right of the text, which looks very subtle and is independent of current or future themes

elstoc: hmm we were almost there as well. We should stick to icons as they're presented in the UI using the default theme for now, to be consistent with both the application and the rest of the documentation. If you want to move to a different standard please raise an issue for discussion and let's do it everywhere. Also, the current convention when referencing icons in a definition list is to put the icon first. Please see https://darktable-org.github.io/dtdocs/module-reference/processing-modules/retouch/ for an example.

quovadit: thanks for the input, I have created an issue (#145)
To follow the current convention I will change it again to the dark background for now.
If the issue is aggreed upon I will have to change all icons anyway.

quovadit:
replaced by new clear pr

PR #143, opened 2020-11-21
Add float TIFFs to supported HDR formats

kmilos:
NB: 16-bit half float native support was added for dt 3.4 recently


elstoc:
Seems a bit odd to mention tiff in both the ldr and sdr sections. Perhaps it would be better if we removed the distinction and just had a single section for "other file formats". 

So just add the sentence in the "hdr file formats" section to the end of the "ldr" section and rename the header, then talk about TIFFs just the once.

elstoc:
I've amended this manually.

kmilos:
Thanks, probably for the best, I wasn't 100% certain about what you intended...

PR #146, opened 2020-11-21
add a color pipeline entry

aurelienpierre:


elstoc:
Thanks @aurelienpierre. I'll merge and do some copy-editing then let you know when I'm done - just so you can do a final sanity check to ensure I haven't changed meanings anywhere.

PR #147, opened 2020-11-21
Suggestions and corrections for overview section

quovadit:
clean version of  #142


elstoc:
Looks fine now @quovadit. There's still the issue with the mouse actions in the filmstrip but I'll fix that up after I merge.

quovadit:
thanks @elstoc

PR #125, opened 2020-11-15
Screen layout with visually marked areas

quovadit:
First of all: I love the new documentation, thanks a lot !!!

For me it's hard to identify the different areas described in Overview / user interface / screen layout 

That's why I edited the image screen-layout.png and marked the 7 areas with colored borders

Just an idea...

Issue #126, opened 2020-11-15
Make a style and provide the tooling to make numbered callouts to annotate images


paperdigits:
Make a style and provide the tooling to make numbered callouts to annotate images

paperdigits:
See something like https://github.com/flameshot-org/flameshot

PR #148, opened 2020-11-21
Screen layout with visually marked areas

quovadit:
clean version of #125, gimp included


elstoc:
@paperdigits: I'm tempted to merge this one. It's clearer than it was before and we can always go back and change it later if need be (we have #126 to discuss it on so it won't get forgotten). What do you think?

paperdigits:
If you think its OK, then merge it.  It once you open the floodgates, the flood comes !)

elstoc:
@paperdigits ok we'll merge this now and you can come up with something better later :)

PR #153, opened 2020-11-22
suggestions and corrections for lighttable section

quovadit:


elstoc:
Some good improvements there. Thanks @quovadit

PR #155, opened 2020-11-22
add info for color calibration

aurelienpierre:


elstoc:
I'm not enormously happy with the amount of duplication in this manual page at the moment - currently we have the explanatory sections for each tab, followed by the module controls for each tab. This means that the manual page jumps about a bit and some information is duplicated between these sections. So I think I'll merge this one, copy-edit, and then try to combine the explanatory and controls sections together somehow.

Job for tomorrow though I think.

aurelienpierre:
The problem is the flat settings explanation doesn't give the workflow tips, implications and interactions with other bits of the software. It's really not enough context. 

But then, the flat settings list should provide a minimal set of workable explanations for people who may not need to go through the whole explanation, for example if they already read it and need a quick reference.

Anyway, that's what I tried to do.

elstoc:
I certainly see the need for a workflow section but it would be better placed closer to the module controls section, for each tab. I'll have a play and see what I can come up with.

PR #164, opened 2020-11-29
Updated slideshow section

quovadit:
updated slideshow section to current behavior (+/-/tab don't work for example)

### DIFF
--- a/content/slideshow/overview.md
+++ b/content/slideshow/overview.md
@@ -5,8 +5,8 @@ weight: 10
 draft: false
 author: "people"
 ---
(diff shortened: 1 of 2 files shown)


elstoc:
@quovadit is this updated to match current _behaviour_ or current _design_? I don't use slideshow but I suspect it's entirely possible there are bugs and it might still be _intended_ to work as advertised in the docs.

quovadit:
as mentioned in the chat: I'll dive deeper into this, try to fix bugs and then adapt the docs to the intended behavior

quovadit:
replaced by #167

PR #167, opened 2020-12-01
Updated slideshow section

quovadit:
2nd attempt to update slideshow section (replaces #164)

Thanks to TurboGit and elstoc the limitations of the slideshow have been fixed (+/- keys) or clarified (left/right shift-key) - see https://github.com/darktable-org/darktable/issues/7128.

So I just removed the misleading wording 'Entering ... will start the slideshow', the not-existing preference-setting and the tab key which is not necessary anymore because panels are hidden automatically.

I moved the F11 key to the Hint-Section at the bottom, because it's no slideshow-specific function.

Finally I added a reference to the similar sticky-preview-feature and put the most-important spacebar and ESC to the top.

I hope that's ok - counter-proposal welcome!

### DIFF
--- a/content/slideshow/overview.md
+++ b/content/slideshow/overview.md
@@ -5,8 +5,8 @@ weight: 10
 draft: false
 author: "people"
 ---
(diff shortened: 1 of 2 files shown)


elstoc:
Looks good. Thanks @quovadit

PR #179, opened 2020-12-16
Update submodule information

cjelgin:
The submodules included in building the site have changed. These changes should reflect the current implementation.


paperdigits:
@cjelgin good catch, thanks!

PR #188, opened 2020-12-30
histogram profile: clarify when it is used

dtorop:
The global color picker entry was out of date. At least as of v3.4 it works in histogram profile, not display.

The histogram entry incorrectly stated that waveform and RGB parade occur in Adobe RGB or display profile. At least as of v3.4 they work in histogram profile.

The clipping warning is calculated in histogram color space, not output color space, at least as of v3.4.

Also clarify that, as these readouts receive data in display colorspace, they may lose some data in the conversion to histogram profile.

Note there's some inconsistency in terminology ("histogram color space" vs. "histogram profile") between the entries. I haven't corrected this, just kept each one internally consistent.

See darktable-org/darktable#3271 for discussion of this.


elstoc:
Looks good. Thanks @dtorop.

PR #189, opened 2020-12-30
French translation

ImagesLD2503:


paperdigits:
If you can hang tight, we are planning to use weblate to translate the docs.

PR #192, opened 2021-01-06
Update description to explain why module is called surface-blur.

matt-maguire:
There were some grumblings on the mail lists why this module was renamed, update the description to better describe the recommended use cases for this module.


elstoc:
Thanks

PR #195, opened 2021-01-18
add a "known issues" section with history

aurelienpierre:


content/special-topics/known-issues.md
```diff
@@ -0,0 +1,46 @@
...
+
+When no entry is detected for a module mandatorily enabled by default, darktable 3.0 and later will initialize a new entry with the default parameters of the module as of the current darktable version.
+
+As a result, edits created with versions prior to 3.0 will lack some history entries which may be initialized by newer versions with different defaults than the ones they were created with, especially if you use the *scene-referred* workflow and the *modern* chromatic adaptation, and the look may change.
+
+## Solution
+
+When you open pictures edited with darktable 2.6 or before:
+
+1. check that the *color calibration* module is turned off. If it is on, it means the *white balance* module left no history entry, which means it was used with "as shot in camera" setting, fetching the white balance setting defined in the RAW metadata. In that case:
```
elstoc, line 29: I don't think that's true. My recollection is that color calibration will only ever be automatically switched on for new edits (those never opened in the darkroom before). The issue with older edits (where the white balance was left unchanged) is that color calibration is *off* and white balance is automatically initialised to D65. So the fix is _either_ to change the WB module to "as shot" _or_ to enable color calibration, depending on whether you want to preserve the old edit or not.

content/special-topics/known-issues.md
```diff
@@ -0,0 +1,46 @@
...
+
+Although darktable aims at keeping compatibility with old edits, there are some limitations that you need to be aware of and cannot be solved in a programmatic way.
+
+# History inconsistencies with modules enabled by default
+
+## Context
+
+The editing history of pictures processed in darktable is saved in the database and in the sidecar XMP files stored in regard of your RAW files, in the same folder. This history records every step of settings change that happened during editing and let you go back to any step at any time.
+
+Before darktable 3.0, modules like *white balance*, *demosaic* and *base curve* that were enabled by default for technical reasons did not write an entry in the editing history if the user did not change the default settings. The software always assumed that no history meant default settings were used. This is not safe because it assumes defaults will never change in the future and it prevents a systematic handling of modules.
```
elstoc, line 17: Minor note. You can use either but it's kind of become a convention that we use underscore for italic.

content/special-topics/known-issues.md
```diff
@@ -0,0 +1,46 @@
...
+## Solution
+
+When you open pictures edited with darktable 2.6 or before:
+
+1. check that the *color calibration* module is turned off. If it is on, it means the *white balance* module left no history entry, which means it was used with "as shot in camera" setting, fetching the white balance setting defined in the RAW metadata. In that case:
+
+    1. disable *color calibration* module,
+    2. reset *white balance* module to "as shot in camera".
+
+2. check that the *filmic rgb* module is turned off. If it is on, it means the *base curve* module left no history entry, which means it was used with the default curve for your camera. In that case:
```
elstoc, line 34: Similarly my understanding is that filmic and base curve will only ever be auto-applied to new edits. See https://github.com/darktable-org/darktable/blob/c1a75a6ac65ea1a7cabf24d940d0fddbf2ecf5b9/src/develop/develop.c#L1391

content/special-topics/known-issues.md
```diff
@@ -0,0 +1,46 @@
...
+
+    1. disable *filmic rgb* module
+    2. turn the *base curve* module on and either reset it or chose the preset corresponding to your camera body.
+
+3. check that the *sharpen* module is turned on. If it is off, search for an entry in the history that might have explicitely disabled it. If there is none, turn it back on.
+
+4. preventively reset the module order to the "legacy" preset.
+
+
+These steps should fix most issues in a safe way. In case you still see garbled output, wrong colors or obvious issues **and** you don't want to troubleshoot the problem **and** you do not care about preserving the original edit, a blind and brute-force way to fix most issues is to compress the history, which will keep what can be kept and reset the rest to sane defaults. But, doing so, you might not be able to recover the original edit, so backup your XMP files first.
```
elstoc, line 44: You mean **discard** history, not **compress**, right? I'm not sure compress will reset anything will it?

aurelienpierre: No, I mean compress. Compressing will fill the voids and keep the rest.

content/special-topics/known-issues.md
aurelienpierre: I need to find that bug again, but I seem to recall it involved both.

content/special-topics/known-issues.md
aurelienpierre: There have been several reports of filmic + base curve over the years, even if it might be fixed somehow, I think it's only safe to ask users to double check manually.

content/special-topics/known-issues.md
elstoc: I've been experiencing this issue on a number of old edits fairly recently

content/special-topics/known-issues.md
elstoc: Ok but can we list it separately because I don't think we've any evidence it's connected to old edits.

content/special-topics/known-issues.md
elstoc: Compressing is just deduplication. It shouldn't change any module's settings. If it does that's a bug.

elstoc:
@aurelienpierre just a reminder that this still needs some more work. Might also be worth looking at the comments in #208.

PR #198, opened 2021-01-19
color calibration : shorten & add caveats

aurelienpierre:
this is refines the explanations with feedback and misunderstanding spotted during the past few weeks on the internets.


elstoc:
Thanks.

PR #201, opened 2021-01-31
color calibration : add color checker

aurelienpierre:


elstoc:
Thanks @aurelienpierre. I'm going to merge this one without doing a detailed review because it's quite long - I might ask you some questions on IRC while I'm copy-editing.

At some point we need to do a general review of some of the more technical modules to see if we can reduce the length or perhaps split into separate pages. They're all getting quite long and technical.

PR #205, opened 2021-02-01
collect: prefs in presets menu

phweyland:
documents:
- not tagged entry #7752
- prefs presets menu entry #7530
- shift double click to switch folder/film roll #7385


elstoc:
Thanks!

PR #206, opened 2021-02-01
metadata view & editor: prefs menu

phweyland:
documents:
- medadata view preferences #7936
- metadata editor preferences #7905


elstoc:
Thanks

PR #212, opened 2021-02-08
Make the Basic Steps section into a task.

paperdigits:
Quick refactor of the Basic Steps section in Overview to give it step numbers. I still think we're giving a bit too much explanation, but I'll take an editing pass at a later time.


elstoc:
Looks good. Thanks.

PR #215, opened 2021-02-09
Move ATI brand out of the spotlight

yuri1969:
It's been over 11 years since the brand discontinuation.


elstoc:
Nice spot. Thanks

PR #216, opened 2021-02-11
geottaging on map

phweyland:
documents #8106


elstoc:
Thanks @phweyland. Are you happy to keep this as two separate sections or would we be better off combining them together under `utility-modules/shared`?

phweyland:
> Thanks @phweyland. Are you happy to keep this as two separate sections or would we be better off combining them together under `utility-modules/shared`?

I'm happy both way ! I just took the easiest way :)

elstoc:
Ok that's fine. I'll move it around when I copy-edit

PR #219, opened 2021-02-13
map: add 3 thumbnail display options

phweyland:
documents #8177

side note: in all files we have applicable-ver**is**on instead of applicable-ver**si**on. Maybe not done on purpose.

### DIFF
--- a/content/module-reference/utility-modules/map/map-settings.md
+++ b/content/module-reference/utility-modules/map/map-settings.md
@@ -1,8 +1,8 @@
 ---
 title: map settings
 id: map-settings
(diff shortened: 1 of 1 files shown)


elstoc:
> side note: in all files we have applicable-verison instead of applicable-version. Maybe not done on purpose.

Thanks I've fixed that now, but now you have a conflict :)

phweyland:
> Thanks I've fixed that now, but now you have a conflict :)

:+1:

elstoc:
Thanks @phweyland

PR #221, opened 2021-02-14
Translation

paperdigits:
Get some scripts and a workflow together for converting all the markdown files to PO.

The current scripts, `tools/generate-translations.sh` creates a temporary po4a config file and then runs po4a.

This addresses #84

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -26,6 +26,12 @@ jobs:
       - name: Build
         run: hugo
 
+      - name: Install po4a
+        run: sudo apt updadte && sudo apt install po4a
+        
+      - name: Generate translated files
+        run: cd tools && bash generate-translations.sh
+        
       - name: Install weasyprint
         run: sudo apt update && sudo apt install weasyprint
 
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -28,10 +28,10 @@ jobs:
 
       - name: Install po4a
         run: sudo apt updadte && sudo apt install po4a
-        
+
       - name: Generate translated files
-        run: cd tools && bash generate-translations.sh
-        
+        run: cd tools && bash generate-translations.sh --no-update
+
       - name: Install weasyprint
         run: sudo apt update && sudo apt install weasyprint
 
@@ -43,4 +43,4 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./public
-          force_orphan: true 
+          force_orphan: true
--- a/config-pdf.yaml
+++ b/config-pdf.yaml
@@ -1,6 +1,9 @@
 baseURL:  https://darktable-org.github.io/dtdocs/
-languageCode: en-us
-title: darktable 3.4 user manual
+defaultContentLanguage: en-us
+languages:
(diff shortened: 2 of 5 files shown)


paperdigits:
The current script almost works, but a few options for po4a that are listed in the man page are flagged by po4a as being not working. Also need to figure out how to parse the yaml fields in the post headers correctly.

elstoc:
Is this likely to work on the gh-pages site or will we need to tweak it (e.g. the current gh-pages doesn't include `/en` in its url)?

paperdigits:
It should work on gh-pages.

elstoc:
@paperdigits please can you also include some documentation of the translation process either in the readme or in contributing.md.

paperdigits:
I think this is ready for prime time. Please test the instructions I've written.

elstoc:
Could we have a link to the translation section from the front page and/or the contributing page please

jpellegrini:
In `translating.md`, there should be a `cd ..` between steps 4 and 5 in section "Generating a PO file". No big deal anyway...

jpellegrini:
I tested this translating workflow for `.po` files, and it worked perfectly!
Thanks to all developers who took the time to make it possible to work with `.po` files.
I'll start translating as soon as this PR is merged.

elstoc:
> Thanks to all developers who took the time to make it possible to work with .po files.

All the credit goes to @paperdigits

elstoc:
How do the weblate translations find their way back into the repository?

jpellegrini:
> How do the weblate translations find their way back into the repository?

Good point. And another question: if any of the translated languages have a translation team, is it OK to have part of the work done via weblate and part directly through .po file editing?  (Not planning to do it; just thought it would be good to tell people if it's possible or not)

paperdigits:
When the translation in weblate hits 100%, weblate will issue a PR against our repo. I don't think it is possible to work in weblate and out side of weblate at the same time. But I'm not a weblate expert, so I might be wrong.

elstoc:
Reading the docs it looks like you can export and import from weblate, so the best thing might be to do everything via weblate, and those that don't want to use the web interface can export/import PO files from there. Otherwise we're going to have fun trying to keep stuff aligned.

jpellegrini:
> Reading the docs it looks like you can export and import from weblate, so the best thing might be to do everything via weblate, and those that don't want to use the web interface can export/import PO files from there. Otherwise we're going to have fun trying to keep stuff aligned.

Ok - but if committing partial translations is not a good thing, then maybe it would be nice to add that to the instructions.

elstoc:
I think that was more about deciding when the translations from weblate get incorporated into the dtdocs repository, and presumably that sort of stuff is going to be configurable. All this will take some trial-and-error to get right.

paperdigits:
I think I was wrong. It seems like if we use weblate, then weblate will manage the translations 100%. You can download the PO files from weblate, then upload them when you're done. Weblate has a cli tool as well, if that is what you like.

paperdigits:
For an easy first project, I added the yaml strings of the hugo PDF theme we use to generate the PDF: https://weblate.pixls.us/projects/darktable/hugo-darktable-docs-pdf-theme/

If someone wants to try and translate those 8 or so strings, please do.

paperdigits:
Closes #84

Issue #222, opened 2021-02-15
PDF build errors


elstoc:
The PDF build CI is showing some errors relating to missing anchor links. I assume this is a change to the website theme that just needs to be ported to the pdf theme.

```
INFO: Step 6 - Drawing
ERROR: No anchor #darktable_user_manual.pdf for internal URI reference
ERROR: No anchor ##white-balance-in-the-chromatic-adaptation-transformation-cat-tab for internal URI reference
ERROR: No anchor ##channel-mixing for internal URI reference
ERROR: No anchor ##brightness-and-colorfulness-tabs for internal URI reference
ERROR: No anchor ##grey-tab for internal URI reference
ERROR: No anchor ##extracting-settings-using-a-color-checker for internal URI reference
ERROR: No anchor #aaec68d6edb7b7b400fdd904228436af#creating-and-editing-presets for internal URI reference
ERROR: No anchor #b5b712dd5ff24196e1d7208044753ee0#spline-controls for internal URI reference
ERROR: No anchor #4dfbb93f42f125092f1f06a38dacabcf#xmp for internal URI reference
ERROR: No anchor #aaec68d6edb7b7b400fdd904228436af#creating-and-editing-presets for internal URI reference
ERROR: No anchor #183604b76b8799b1634ce41712504cfa#notes for internal URI reference
INFO: Step 7 - Adding PDF metadata
```

@paperdigits could you take a look please

paperdigits:
I've fixed everything except the first error, which is the PDF linking to itself. This doesn't make any sense, so we should do something about the PDF links. Either move that link into the website template or something else. I'm open to suggestions.

elstoc:
We'll leave this open to track the remaining PDF link error

elstoc:
For the PDF issue, we could wrap the paragraph in a `<div class="webonly">` and just hide it in the PDF theme. CSS

paperdigits:
Now you're thinking like a champion!

elstoc:
Resolved by https://github.com/darktable-org/dtdocs/commit/31c5158542d42ab66964023381f472c83e4c3cf1

PR #223, opened 2021-02-17
Fixes links targeting a heading in the same file. 

paperdigits:
This fixes all errors in #222 except for the first error, which is the link to the PDF. Since the PDF really shouldn't like to itself, we should probably put the link to the PDF in the template for the website. Or something else, I'm open to suggestions.


elstoc:
Thanks @paperdigits. I'm not greatly bothered by the pdf link for the moment, so let's merge this now and think about that later.

PR #224, opened 2021-02-20
blend-modes: complete documentation

### LINKED ISSUE #156: better explain blend modes
Amend `darkroom/masking-and-blending/blend-modes.md` to include explanations of all available blend modes, including those introduced with the scene-referred RGB blending changes.

elstoc:
Resolves #156

### DIFF
--- a/content/darkroom/masking-and-blending/blend-modes.md
+++ b/content/darkroom/masking-and-blending/blend-modes.md
@@ -5,44 +5,163 @@ weight: 20
 draft: false
 ---
 
-Blend modes define how the input and output of a module are combined (blended) together before a module's final output is passed to the next module in the pixelpipe. Most of the blend modes should be familiar from other imaging software and are thoroughly discussed in the [GIMP user manual](https://docs.gimp.org/2.8/en/gimp-concepts-layer-modes.html). 
+Blend modes define how the input and output of a module are combined (blended) together before a module's final output is passed to the next module in the pixelpipe. 
 
-This section therefore only discusses some of the darktable-specific blend modes in detail.
+Classic blending modes, designed for display-referred RGB (constrained to 0-100%) implicitly define a fulcrum at 50% (grey) or 100% (white) in their algorithms, depending on the blend mode. Because scene-referred is not subject to these restrictions, this fulcrum needs to be explicitly defined by the user when performing blending operations in the "RGB (scene)" color space. The additional _blend fulcrum_ parameter will be presented to the user when using one of these blend modes in this color space. The effect depends on the operator used. For example, values above the fulcrum might be brightened and values below darkened, or vice versa.
 
-Some blend modes depend on a fulcrum for their operation. This fulcrum usually defines the point at which the blend mode results in a "no-operation" which may be white or grey depending on the blend mode. The additional _blend fulcrum_ parameter allows the position of this fulcrum to be adjusted where blending is performed within the RGB scene-referred color space. The effect depends on the operator used. For example, values above the fulcrum might be brightened and values below darkened, or vice versa.
+The final output of a module is computed 'per-pixel' as follows:
+
+```
+final_output = (1.0 - opacity) * module_input + opacity * blended_output
+```
+
+where the `blended_output` is a combination of the input and output images, depending on the blend mode (below), and the `opacity` is defined 'per-pixel' by a combination of the mask and global opacity parameter.
+
+# normal modes
 
 normal
-: The most commonly used blend mode, "normal" simply mixes input and output to an extent determined by the opacity parameter. This mode is commonly used to reduce the strength of a module's effect by reducing the opacity. This is also usually the blend mode of choice when applying a module's effect selectively with masks.
+: The most commonly used blend mode, "normal" simply mixes input and output to an extent determined by the opacity parameter. This mode is commonly used to reduce the strength of a module's effect by reducing the opacity. This is also usually the blend mode of choice when applying a module's effect selectively with masks. This mode is also known as the "over" Porter-Duff alpha blending operator (see [Alpha compositing](https://en.wikipedia.org/wiki/Alpha_compositing) for more details).
 
 normal bounded
+: _not available in the "RGB (scene)" color space_
 : This blend mode is the same as “normal”, except that the input and output data are clamped to a particular min/max value range. Out-of-range values are effectively blocked and are not passed to subsequent modules. Sometimes this helps to prevent artifacts. However, in most cases (e.g. highly color-saturated extreme highlights) it is better to let unbound values travel through the pixelpipe to be properly handled later. The “normal” blend mode is therefore usually preferred.
 
-lightness
-: This blend mode mixes lightness from the input and output images. Color (chroma and hue) is taken unaltered from the input image.
+# arithmetic modes
 
-chroma
-: This blend mode mixes chroma (saturation) from the input and output images. Lightness and hue are taken unaltered from the input image.
+addition
+: Add together the pixel values of the input and output images, lightening the output. When blending in the "RGB (scene)" color space, the pixel values of the output image are multiplied by a value proportional to the "blend fulcrum".
+
+subtract
+: Subtract the pixel value of the _output_ from the _input_. When blending in the "RGB (scene)" color space, the pixel values of the output image are multiplied by a value proportional to the "blend fulcrum". Pixel values less than 0 are set to 0.
+
+subtract reverse
+: _only available in the "RGB (scene)" color space_
+: Subtract the pixel value of the _input_ from the _output_. In this case, the "blend fulcrum" parameter of the "RGB (scene)" color space is applied to the input image.
+
+multiply
+: Multiply the pixel values of the input and output together. When blending in display-referred color spaces, pixel values are between 0 and 1.0, the final output will be clamped and will always be darker. When blending in the "RGB (scene)" color space, this value is further multiplied by a value proportional to the "blend fulcrum". In this case, values may be greater than 1.0 and therefore brighten the base image. This may have other side-effects, such as updating the white point in the filmic module.
+
+: Multiply blending simulates an optical variable density filter, where the density is defined by the output of the module. It has many applications, from blooming and local contrast enhancements (when used with a blur or low-pass filter) to dodging/burning and global contrast enhancements (when used with exposure). The fulcrum sets the output intensity threshold between darkening and brightening (any RGB value below fulcrum will darken).
+
+multiply reverse
+: _only available in the "RGB (scene)" color space_
+: This mode works in the same way as the "multiply" mode except that in the `final_output` equation given at the top of this page, the `module_input` parameter is replaced with `module_output` (the output of the module before blending is applied).
+
+divide
+: Divide the pixel values of the input by the output. When blending in the "RGB (scene)" color space, the pixel values of the output image are multiplied by a value proportional to the "blend fulcrum".
+
+: Since this is the inverse of the multiply mode, it will darken where multiply brightens and vice versa. Everything else works in essentially the same way.
+
+divide reverse
+: _only available in the "RGB (scene)" color space_
+: Divide the pixel values of the _output_ by the _input_. In this case, the "blend fulcrum" parameter of the "RGB (scene)" color space is applied to the input image.
+
+screen
+: _not available in the "RGB (scene)" color space_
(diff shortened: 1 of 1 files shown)


elstoc:
@hlecleme I think you've done some work in this area. Would you be able to review this please?

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
 draft: false
 ---
 
-Blend modes define how the input and output of a module are combined (blended) together before a module's final output is passed to the next module in the pixelpipe. Most of the blend modes should be familiar from other imaging software and are thoroughly discussed in the [GIMP user manual](https://docs.gimp.org/2.8/en/gimp-concepts-layer-modes.html). 
-
-This section therefore only discusses some of the darktable-specific blend modes in detail.
+Blend modes define how the input and output of a module are combined (blended) together before a module's final output is passed to the next module in the pixelpipe. 
 
 Some blend modes depend on a fulcrum for their operation. This fulcrum usually defines the point at which the blend mode results in a "no-operation" which may be white or grey depending on the blend mode. The additional _blend fulcrum_ parameter allows the position of this fulcrum to be adjusted where blending is performed within the RGB scene-referred color space. The effect depends on the operator used. For example, values above the fulcrum might be brightened and values below darkened, or vice versa.
```
aurelienpierre, line 10: I would add here : "Classic blending modes, meant for display-referred RGB (forcefully encoded between 0 and 100 %), implicitly define a fulcrum at 50 % (grey) or 100 % (white) in the code. Because scene-referred is not bounded to 100 % anymore, that fulcrum needs to be explicitely defined by users.".

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+```
+final_output = (1.0 - opacity) * module_input + opacity * blended_output
+```
+
+where the `blended_output` is a combination of the input and output images, depending on the blend mode (below), and the `opacity` is defined 'per-pixel' by a combination of the mask and global opacity parameter.
+
+# normal modes
+
 normal
 : The most commonly used blend mode, "normal" simply mixes input and output to an extent determined by the opacity parameter. This mode is commonly used to reduce the strength of a module's effect by reducing the opacity. This is also usually the blend mode of choice when applying a module's effect selectively with masks.
```
aurelienpierre, line 23: Add : "The normal mode is the "over" Porter-Duff alpha blending operator."

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
 
-color
-: This blend mode mixes color (chroma and hue) from the input and output images. Lightness is taken unaltered from the input image. 
+subtract reverse
+: _only available in the "RGB (Scene)" color space_
+: Subtract the pixel value of the _input_ from the _output_. In this case, the "blend fulcrum" parameter of the "RGB (scene)" color space is applied to the input image.
 
-: _Caution: When modules drastically modify hue (e.g. when generating complementary colors) this blend mode can result in strong color noise._
+multiply
+: Multiply the pixel values of the input and output together. As all pixel values are between 0 and 1.0, the resulting output will always be darker. When blending in the "RGB (scene)" color space, this value is further multiplied by a value proportional to the "blend fulcrum".
```
aurelienpierre, line 42: You may add : "Multiply simulates an optical variable density filter, where the density is defined by the output of the module. It has many application, from blooming and local contrast enhancements (used with a blur or low-pass filter output) to dodging/burning and global contrast enhancements (used with exposure output). The fulcrum sets the output intensity threshold between darkening and brightening (any RGB value below fulcrum will darken)".

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+
+multiply reverse
+: _only available in the "RGB (Scene)" color space_
+: This mode works in the same way as the "multiply" mode except that in the `final_output` equation given at the top of this page, the `input` parameter is replaced with `module_output` (the output of the module before blending is applied).
+
+divide
+: Divide the pixel values of the input by the output. When blending in the "RGB (scene)" color space, the pixel values of the output image are multiplied by a value proportional to the "blend fulcrum".
```
aurelienpierre, line 49: You may add "Since this is the inverse of the multiply mode, it will darken where multiply brightens, and the other way around. Everything else works essentially the same.".

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+
+divide reverse
+: _only available in the "RGB (Scene)" color space_
+: Divide the pixel values of the _output_ by the _input_. In this case, the "blend fulcrum" parameter of the "RGB (scene)" color space is applied to the input image.
+
+screen
+: _not available in the "RGB (Scene)" color space_
+: Invert the input and output pixel values, multiply those values together and invert the result. This yields the opposite effect to "multiply" mode -- the resulting image is usually brighter, and sometimes “washed out” in appearance.
```
aurelienpierre, line 57: replace "this is the opposite" by "this is approximately the opposite", because it is obviously not arithmetically equivalent and also assumes non-radiometrically-linear RGB.

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+
+average
```
aurelienpierre, line 59: the arithmetic mean (because we later have geometric and harmonic).

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+
+geometric mean
+: Return the square root of the product of the input and output pixel values.
+
+harmonic mean
+: Return the product of the input and output pixel values, multiplied by 2 and divided by their sum.
+
+# contrast enhancing modes
+
+The following modes are not available in the "RGB (Scene)" blending color space as they rely on an assumption of "50% mid grey" which only applies to display-referred color spaces.
```
aurelienpierre, line 73: display-referred AND non-linear, because display-referred only means the signal is bounded in [0; 100] % but says nothing about the encoding.

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+: Mix the "green" channel from the input and output images, while taking the other channels unaltered from the input image.
+
+RGB blue channel
+: Mix the "blue" channel from the input and output images, while taking the other channels unaltered from the input image.
+
+# HSV channels
+
+The following are available when blending in RGB color spaces only.
 
 HSV lightness
```
aurelienpierre, line 128: These are HSL, not HSV (V = max(RGB), L = (min(RGB) + max(RGB)) / 2).

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
-: Only available with modules that work in the RGB color space, this blend mode mixes lightness from the input and output images, while color is taken only from the input image. In contrast to “lightness” this blend mode does not involve clamping.
+: Mix the lightness from the input and output images, while taking color unaltered from the input image. In contrast to “lightness” this blend mode does not involve clamping.
 
 HSV color
-: Only available with modules that work in the RGB color space, this blend mode mixes color from the input and output images, while lightness is taken only from the input image. In contrast to “color” this blend mode does not involve clamping.
+: Mix the color from the input and output images, while taking lightness unaltered from the input image. In contrast to “color” this blend mode does not involve clamping.
```
aurelienpierre, line 132: HSL color is not suitable for use with scene-referred images where some pixels have RGB values above 1. For such pixels, the saturation value (S) will become negative as a result of the way the lightness (L) is computed. This is not a bug but a property of the HSL decomposition. For such pixels, darktable sets H = S = 0 as a fall-back so the color blending may not work as expected.

content/darkroom/masking-and-blending/blend-modes.md
```diff
@@ -5,44 +5,159 @@ weight: 20
...
+
+# others
+
+lightness
+: Mix lightness from the input and output images, while taking color (chroma and hue) unaltered from the input image.
+
+chroma
+: Mix chroma (saturation) from the input and output images, while taking lightness and hue unaltered from the input image.
```
aurelienpierre, line 140: it's actually chromaticity

content/darkroom/masking-and-blending/blend-modes.md
elstoc: can you clarify what you'd like me to change here please.

content/darkroom/masking-and-blending/blend-modes.md
elstoc: ok but they're all **called** HSV so what do you want me to say here?

content/darkroom/masking-and-blending/blend-modes.md
elstoc: ok then how does this differ from "color"?

content/darkroom/masking-and-blending/blend-modes.md
elstoc: Again giving something a label of HSV and then calling it HSL in the docs is confusing.

content/darkroom/masking-and-blending/blend-modes.md
aurelienpierre: They should all be called HSL.

content/darkroom/masking-and-blending/blend-modes.md
aurelienpierre: Then the HSV label is wrong in GUI and should be changed.

content/darkroom/masking-and-blending/blend-modes.md
aurelienpierre: "color" uses H and S from HSL (so, expects display-referred RGB). Chroma uses RGB ratios, divided by an euclidean norm, like filmic, color calibration etc. and doesn't care about bounds.

content/darkroom/masking-and-blending/blend-modes.md
aurelienpierre: > Edit: or should it just be "...only applies to display-referred and non-linear color spaces."

yes

content/darkroom/masking-and-blending/blend-modes.md
aurelienpierre: Forget about that, sorry for the noise.

content/darkroom/masking-and-blending/blend-modes.md
aurelienpierre: Ditch that too.

content/darkroom/masking-and-blending/blend-modes.md
elstoc: did you want the name to be changed on this as well (in darktable) or just the description?

content/darkroom/masking-and-blending/blend-modes.md
matt-maguire: what is meant here by "invert"? I presume it can't mean taking the reciprocal, since 1/((1/a)\*(1/b)) = a\*b, and the description stayed this gives a different result to multiplication.

I can guess it might be something like 1 - (1-a)*(1-b), but the word "invert" is really ambiguous. this may be a good example justifying the inclusion of formulas in the descriptions.

elstoc: I copied a lot of this stuff from the GIMP documentation and other stuff I found.

hlecleme (commented):
Thanks for the work. Globally it looks good to me.

elstoc:
Thanks @hlecleme

elstoc:
I've made some amendments based on the feedback. If there are no objections I'll merge this tomorrow. We can always change it again later if needed.

content/darkroom/masking-and-blending/blend-modes.md
elstoc: I've just changed the description now.

content/darkroom/masking-and-blending/blend-modes.md
elstoc: I've made some changes from your suggestions but I'm still not 100% certain what you wanted me to write.

elstoc:
Ok merging now. Let me know if you'd like any further changes.

PR #226, opened 2021-02-21
tagging: attempt to clarify what are the tags in darktable

phweyland:
@quovadit I'm very aware of the role of the documentation within the difficulties we had to understand each other.

The usage of the word category is particularly confusing.

Do the changes made improve a bit the vision of what are tags in dt ? TIA.

### DIFF
--- a/content/module-reference/utility-modules/shared/tagging.md
+++ b/content/module-reference/utility-modules/shared/tagging.md
@@ -2,29 +2,64 @@
 title: tagging
 id: tagging
 applicable-version: 3.6
-tags: 
+tags:
 view: lighttable, [darkroom], tethering, map
 ---
 
 Manage tags attached to images.
 
-Tags provide a means of adding information to images using a keyword dictionary. You can also manage tags as a hierarchical tree, which can be useful when their number becomes large. Only tags at the leaves of the tree add information to an image -- parent tags simply serve as categories and are are used to organise the leaf tags. For example, in the “color|white” tag, “color” is a category for the “white” tag.
+Tags provide a means of adding information to images using a keyword dictionary. You can also manage tags as a hierarchical tree, which can be useful when their number becomes large.
 
 Tags are physically stored in [XMP sidecar files](../../../overview/sidecar-files/_index.md) as well as in darktable's library database and can be included in [exported](./export.md) images.
+
+# definitions
+
+The following definitions assume that you have set up a single tag named `places|France|Nord|Lille`.
+
+tag
+: A tag is a descriptive string that may be attached to an image. A tag can be a single term or a sequence of connected terms forming a path, separated by the pipe symbol. For example, `places|France|Nord|Lille` defines a single tag, with each term in the path forming a subset of those before it. You can attach as many tags to an image as you like.
+
+You can assign properties (name, private, category, synonyms and image order) to a tag.
+
+node
+: Any path that forms part of a tag is a node. In the above example, `places`, `places|France`, `places|France|Nord` and `places|France|Nord|Lille` are all nodes.
+
+free node
+: Any node that is not explicitly defined as a tag is called a free node. In the above example, `places`, `places|France` and `places|France|Nord` are all free nodes. You cannot set any properties, except “name”, for a free node and you cannot add a free node to an image. See the “multiple tags” section below for more information.
+
+category
+: Any tag can be flagged by the user as being a “category”.
+
+# multiple tags
+
+The above definitions considered a simple example – a single tag and its properties. Consider instead the scenario where the following tags are each separately defined in darktable.
+
+`places|France|Nord|Lille`
+`places|France|Nord`
+`places|France`
+`places|England|London`
+
+In this case, these are four pipe-delimited tags. The nodes are `places`, `places|France`, `places|France|Nord`, `places|France|Nord|Lille`, `places|England` and `places|England|London`.
+
+The only free nodes are `places` and `places|England`. Both of these free nodes are also (by implication) categories.
+
+You can attach any of these tags to any image. Any tags attached to an image, except category, can be included when that image is exported (see [export](./export.md) module).
+
+If you attach the `places|France|Nord|Lille` tag to an image, the `places|France|Nord` and `places|France tags` are also implicitly defined against that image (you don’t need to attach them yourself). Note that this is only true here because those additional tags have been separately defined – the `places` node is not included because it has not been separately defined as a tag.
 
 # module sections
 
 The tagging module consists of two sections
 
-1. The upper _attached tags_ section (with the _attach/detatch_ buttons under it)
+1. The upper _attached tags_ section (with the _attach/detach_ buttons under it)
 
-2. The lower _tag dictonary_ section (with the _new/import.../export..._ buttons under it)
+2. The lower _tag dictionary_ section (with the _new/import.../export..._ buttons under it)
 
 ![tagging-overview](./tagging-overview/tagging-overview.png#w33)
 
 ## attached tags section
 
-The _attached tags_ section displays tag(s) attached to image(s) that are 
+The _attached tags_ section displays tag(s) attached to image(s) that are
 
 - currently selected (if the mouse cursor is not hovering over an image on the lighttable); or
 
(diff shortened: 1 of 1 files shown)


content/module-reference/utility-modules/shared/tagging.md
```diff
@@ -55,16 +66,16 @@ The _tag dictionary_ section displays all of the tags that are available in dark
...
+- a minus sign [--] indicates that the tag is attached to at least one of the selected images. If the symbol is next to a node name in the hierarchical _tree_ view, it means that one of the child tags under that node is attached to at least one of the selected images.
 
-- if no indicator symbol is present, this means that the tag is not attached to any of the selected images, or that the category has no child tags attached to any of the selected images.
+- if no indicator symbol is present, this means that the tag is not attached to any of the selected images, or that the node has no child tags attached to any of the selected images.
 
-In the hierarchical _tree_ view, a name in italics represents a category, and not a child tag. Note that categories are just used to organise the tags, and that a category cannot be directly attached to an image or exported.
+In the hierarchical _tree_ view, a name in italics represents a category (a node which not a tag or a tag set as category). Note that nodes are just used to organise the tags, and that a node which is not a tag cannot be directly attached to an image.
```
elstoc, line 73: I'm not sure that the text in the brackets is clear. Are you saying that a category is "**either** a node that is not a tag, **or** a tag that has been explicitly defined as a category"? If so, I think it should be stated that way (with both the "either" and the "or" for clarity). 

In the second sentence, where you say "a node which is not a tag" are you just trying to avoid saying the word "category"? or are you explicitly excluding the second definition of "category" from the first sentence?

phweyland: > with both the "either" and the "or" for clarity

I'll do.

> where you say "a node which is not a tag" are you just trying to avoid saying the word "category" or are you explicitly excluding the second definition of "category" from the first sentence ?


I don't how to say another way, but yes, a node ~~which~~ that is not tag (which has also the category property, but the fact it cannot be attached doesn't come from its category property but from the absence of tag).

>  The phrase "a node that is not a tag" defines a particular type of node (a subset of all nodes).

That's the case. I'll use "that" instead of "which" then.

Thanks

elstoc: > I don't how to say another way, but yes, a node which that is not tag (which has also the category property, but the fact it cannot be attached doesn't come from its category property but from the absence of tag).

I'm still not entirely sure I understand. Perhaps I need to do some more reading. It doesn't help that I don't use tagging for my images.

phweyland: I've tried to use properly which and that ... not yet completely sure though  :)
I've reshaped a bit the critical paragraph. I hope that's better now.

elstoc:
Perhaps we need to consider a shorter name for "node-that-is-not-a-tag". We could perhaps use the terms "leaf node" and "branch node" with clear definitions up front?

phweyland:
> Perhaps we need to consider a shorter name for "node-that-is-not-a-tag". We could perhaps use the terms "leaf node" and "branch node" with clear definitions up front?

It would be nice. I think that's the reason why category was used to cover this, but leading to confusion.
Actually there are only two fundamental cases:
- node (that is not a tag) and
- tag

As soon as there is a tag for that node it's a tag (and not anymore a node that is not a tag :) )

So, once properly defined, we could keep "node" and "tag" only.

"Leaf" is related to the hierarchy. A leaf is always a tag. But a tag is not always a leaf, as we can have tags along a branch.

"Category" is only a property. A node is always a category. A tag is a category if the user decides so (for example to define synonyms).

elstoc:
> "Category" is only a property. A node is always a category. A tag is a category if the user decides so (for example to define synonyms)

So that's the only thing that's difficult - the fact that you can flag a tag as a category even if it's not currently used as a "node". If we're saying that "a node is not a tag" then we can use those two terms and define them up-front.

phweyland:
> the fact that you can flag a tag as a category even if it's not currently used as a "node". 

Not sure to understand what you mean...
The user creates tags, not nodes. Nodes come with the hierarchy of that tag but they have no specific usage, other than building the tree.

> If we're saying that "a node is not a tag" then we can use those two terms and define them up-front.

This would be only a language convention. Because, from the tree standpoint every tag is on a tree node...

What about "empty node" or "free node" instead of "node that is not a tag" ?

phweyland:
Another confusion I've made :(
A leaf is not node.
Each leaf is tag. But a tag is either a leaf or a node.

phweyland:
Added text from HackMD. Thanks !

content/module-reference/utility-modules/shared/tagging.md
```diff
@@ -2,29 +2,64 @@
...
+# multiple tags
+
+The above definitions considered a simple example – a single tag and its properties. Consider instead the scenario where the following tags are each separately defined in darktable. For now, assume that none of the tags has been flagged as a “category”.
+
+`places|France|Nord|Lille`
+`places|France|Nord`
+`places|France`
+`places|England|London`
+
+In this case, each of these pipe-delimited strings is a tag. The nodes are `places`, `places|France`, `places|France|Nord`, `places|France|Nord|Lille`, `places|England` and `places|England|London`.
```
quovadit, line 42: I would say 'each of these rows' instead of 'each of these pipe-delimited strings'.
Although it's correct, you could think that each part separated by pipes is a tag (places, France, ...)

content/module-reference/utility-modules/shared/tagging.md
```diff
@@ -2,29 +2,64 @@
...
+
+free node
+: Any node that is not explicitly defined as a tag is called a free node. In the above example, `places`, `places|France` and `places|France|Nord` are all free nodes. You cannot set any properties, except “name”, for a free node and you cannot add a free node to an image. See the “multiple tags” section below for more information.
+
+category
+: Any tag can be flagged by the user as being a “category”. All free nodes are implicitly defined as “category” tags.
+
+# multiple tags
+
+The above definitions considered a simple example – a single tag and its properties. Consider instead the scenario where the following tags are each separately defined in darktable. For now, assume that none of the tags has been flagged as a “category”.
```
quovadit, line 35: why is it necessary to exclude tags flagged as category?

content/module-reference/utility-modules/shared/tagging.md
```diff
@@ -2,29 +2,64 @@
...
+`places|France|Nord|Lille`
+`places|France|Nord`
+`places|France`
+`places|England|London`
+
+In this case, each of these pipe-delimited strings is a tag. The nodes are `places`, `places|France`, `places|France|Nord`, `places|France|Nord|Lille`, `places|England` and `places|England|London`.
+
+The only free nodes are `places` and `places|England`. Both of these free nodes are also (by implication) categories.
+
+You can attach any of these tags to any image. Any tags attached to an image (except “category” tags) can be included when that image is exported (see [export](./export.md) module).
```
quovadit, line 46: I would say 'tags flagged as "category"' instead of '"category" tags'

content/module-reference/utility-modules/shared/tagging.md
```diff
@@ -118,17 +153,17 @@ A number of tags are also generated automatically by darktable (e.g. “darktabl
...
 
-- A tag set as “private” is, by default, not exported. 
+- A tag set as “private” is, by default, not exported.
 
 - A tag set as “category” is not exported in `XMP-dc Subject`. However it is exported in `XMP-lr Hierarchical Subject` as this XMP metadata holds the organization of your tags.
 
 - “synonyms” enrich the tag information and are mainly used to assist search engines. For example “juvenile”, “kid” or “youth” can be set as synonyms of “child”. Synonyms can also be used to translate tag names to other languages.
 
 The "rename path..." operation is only available in the _tree_ view mode, and it shows the number of tagged images which would be impacted by a change to the name of this tag or category. The rename path window lets the user change the full path of the tag, including the categories to which it belongs (categories can be specified using the pipe `|` symbol). This operation is powerful, but please take care as it can have a significant impact on the metadata of your images. The operation is aborted if the requested change causes a conflict with an existing tag.
```
quovadit, line 164: Here are 3 'categories' left which should be nodes I think

And 'change the full path of the tag' should be 'change the full path of the node'

quovadit (commented):
I love this version, it's extremely clear now, thank you!

content/module-reference/utility-modules/shared/tagging.md
elstoc: I'm not sure about "rows" (really that word should only be used in the context of tables) but we could say that there are four pipe-delimited tags

content/module-reference/utility-modules/shared/tagging.md
elstoc: I got the impression that category tags are not attached to images. If so, it affects the paragraph below where we talk about implicitly attached tags.

content/module-reference/utility-modules/shared/tagging.md
elstoc: Actually I was going to suggest going the other way. "tags flagged as "category"' is pretty ugly English. Given that we've defined category tags in the first section I'd say "category tags" is a neater phrase to use throughout. As long as we're consistent and have defined our terms well it should be fine.

content/module-reference/utility-modules/shared/tagging.md
```diff
@@ -2,29 +2,64 @@
...
+
+You can assign properties (name, private, category, synonyms and image order) to a tag.
+
+node
+: Any path that forms part of a tag is a node. In the above example, `places`, `places|France`, `places|France|Nord` and `places|France|Nord|Lille` are all nodes.
+
+free node
+: Any node that is not explicitly defined as a tag is called a free node. In the above example, `places`, `places|France` and `places|France|Nord` are all free nodes. You cannot set any properties, except “name”, for a free node and you cannot add a free node to an image. See the “multiple tags” section below for more information.
+
+category
```
elstoc, line 30: to go with my responses to @quovadit below, we should change this to "category tag"

content/module-reference/utility-modules/shared/tagging.md
phweyland: Only free node cannot be attached. All tags, even those flagged as category, can be attached.

elstoc: ok then maybe we can remove that sentence.

phweyland:
I'll update with your comments.
I think also that saying that a free node is as a category is useless and source of confusion. It's enough to say it cannot be attached and or exported. Then a "category" is a tag flagged as category and only that.

elstoc:
ok that makes sense

elstoc:
@phweyland is this complete now?

phweyland:
For me that's fine. If @quovadit and yourself are OK, that's done.
Thank you both !

quovadit:
perfect for me

elstoc:
yep definitely a much needed improvement. I'll give it another clean pass later on for a copy-edit but will merge now.

PR #230, opened 2021-02-25
Use nixpkg in CI

paperdigits:
Switch actions to use nixpkg.


elstoc:
Can we run linkchecker in the gh-pages build as well please, because I'm a cowboy and quite often commit directly to master.

PR #232, opened 2021-03-01
Set as a tag

quovadit:
Update utility-modules/shared/tagging/ according  to PR [#8270](https://github.com/darktable-org/darktable/pull/8270):
- new context menu "set as a tag" 
- omit "delete path" when the tag is a leaf
- rename existing context menu entries


elstoc:
Thanks @quovadit

PR #236, opened 2021-03-05
denoise-profiled: amend autoset parameter description

### LINKED ISSUE #235: noise (profiled): confusing description of "adjust autoset parameters"
The relevant manual section says:

> In general, the value of this slider indicates the number of stops by which you had to increase the exposure. For example, if you had to push the exposure 2 stops above what you normally use, then set this slider to 2.

This confuses me to the point that I do not understand what this phrase is trying to say.

The default value of this setting is 1.  Taking the example of exposure pushed 2 stops above what one normally uses, I could understand if the slider had to be set to 4 (if its scale is linear), or to 3 (if it's log2).  

But 2?  That's consistent with square-root-of-linear, but if this parameter indeed works this way, the documentation should be more explicit.

elstoc:
Resolves #235


elstoc:
@rawfiner @matt-maguire ^^ this?

rawfiner:
perfect for me :-)

elstoc:
ok that's good enough for me. Thanks

PR #240, opened 2021-03-13
[WIP] quickstart tour

### LINKED ISSUE #209: [FR] "getting started" / "Quick Start Tour" section
There has been talk at various times about improving the "onboarding" process for new dt users.  To help in this, the manual should have a section _right_ _at_ _the_ _beginning_ that takes a user through importing an image/folder, making some simple edits, and exporting the result.  While also pointing out powerful features such as presets, shortcuts, etc. with links to the appropriate sections of the manual.

This would also be a good place to link Bruce Williams' recent three-part "dt 3.4 for newbs" video.  New users will never find their way to the resources chapter in the back of the manual....

ralfbrown:
A first draft of a quick-start tour.  I haven't embedded any images yet or done any cross-linking, nor even checked that it renders correctly.  I'll deal with all of that once a more-or-less final text is approved.

The text is deliberately fairly informal in tone.  We want to engage the user rather than have it feel like a dry reference work.

Fixes #209.


paperdigits:
Can you please mark this as a WIP?

paperdigits:
Should this exist under `dtdocs/content/guides-tutorials/`?

ralfbrown:
A quick-start should be at the very beginning of the documentation so that people can't possibly miss it.  When was the last time you saw a quick-start hidden in an appendix?

As I said, formatting/images/links/etc. will be fixed up once I know that this is actually going somewhere.

elstoc:
I agree that the first sections could use some more direction and be more concise but I'm not sure whether we need a quickstart section _in addition to_ the existing overview - there seems to be quite a lot of duplication at the moment. I don't have a clear idea of what the "right" solution is but my initial thoughts are that it should either be something much more brief (with links to the more comprehensive treatments) or should be integrated into the overview section.

ralfbrown:
You also need to keep user behavior in mind -- of the minority who read the documentation at all, only a tiny minority will bother to follow xref links.  So you need to have all of the important information in one sequential narrative.  Plus it's vital to have a sequence of steps the user can follow along with (and maybe try variations) from start to finish -- learning by doing.

The Retouch section could be reduced to a mention and link, but we need to keep the info about ability to type text into bauhaus sliders and the aspect ratio pulldown, auto-apply presets to modify dt defaults (possibly shortened), Control+scroll to resize fields, etc. in the main flow.  Those are the things causing the most user questions that I've seen.

Eventually, there should be a splash screen at startup controlled by a darktablerc setting.  If absent, pop up a dialog with "new to dt?  take the quick-start tour" and a button to take one there.  If present but for a version less than that running, pop up "new to vN.NN?  see what's new" and a button to take one to the release notes or something similar.  And then of course update the darktablerc setting.

paperdigits:
I understand your vision, but share the same concern as @elstoc:

1. The duplication of content between the quick start and the overview section is concerning. Could we change the overview section to better suite your goal?

2. No User Guide for an application with similar complexity to darktable starts with a quick start. The quick start is usually delivered as a separate document.

3. I'd expect this to be very task driven, but it doesn't seem to be. There are no numbered steps. It also fluctuates too freely between conceptual and workflow information. Too much conceptual stuff in the beginning.

4. It doesn't follow out documented conventions.

paperdigits:
Hi @ralfbrown I see you changed the sections to be numbered, but I was referring to have discrete steps in the body text, not section numbering (which I don't really care about). 

Can you perhaps address the issues we've raised?

PR #254, opened 2021-04-18
create a color dimensions page

aurelienpierre:
prepares color balance RGB background


elstoc:
I know you're in the process of getting user feedback on this from various sources (pixls, irc etc.) so I think I'll merge this now and push a quick copy-edit so that it's easier for others to review and feed back.

PR #266, opened 2021-05-09
doc updates

phweyland:
- lut3d: add file formats
- locations: add polygon shape


elstoc:
Thanks @phweyland I'll merge now and do a quick copy-edit

phweyland:
> The format .gmz a compressed format that is only available when GMIC is installed. 

In lut3d. Wouldn't that be better :

The compressed format .gmz is only available when GMIC is installed. 

 ?

PR #272, opened 2021-05-19
document color balance RGB

aurelienpierre:


elstoc:
Thanks @aurelienpierre. I'll merge and copy-edit now, then perhaps it might be good to share it on pixls for some feedback?

PR #275, opened 2021-06-03
Add man page for purge_non_existing_images.sh script.

matt-maguire:
create man page for `purge_non_existing_images.sh` script.

It warns to make a database backup, but it seems there is no procedure for this in the user guide. So, probably yet another topic to tackle.


matt-maguire (commented):
Forgot to update the id in the metatags

elstoc:
Thanks

PR #277, opened 2021-06-05
lens correction: warn the user about potential artefacts at borders

rawfiner:
Currently, unknown pixels at borders are filled with border pixels (previously they were black) to avoid impacting the algorithms later in the pipe that blur the image.
As pointed out in https://github.com/darktable-org/darktable/issues/9090 and in the original PR https://github.com/darktable-org/darktable/pull/8873 it may give some artefacts. As a solution for now, warn the user about these artefacts in the documentation.
Long term solution will probably be to have an autocrop setting in lens correction, but this probably won't land for 3.6


elstoc:
Thanks

PR #278, opened 2021-06-08
update module order with new modules for 3.6

matt-maguire:
update default module order page to include new modules (color calibration, censorize, color balance rgb) and new deprecations.


elstoc:
Thanks

PR #286, opened 2021-06-24
color calibration : improve gamut clipping/mapping explanations

aurelienpierre:


content/module-reference/processing-modules/color-calibration.md
```diff
@@ -112,10 +112,22 @@ chroma
...
+
+In all these cases, you **must** enable the clipping of negative RGB values in the _color calibration_ module.
+
+**Note 2**: A common case for failure of the color algorithms in _color calibration_ (especially the gamut compression) is pixels that have a luminance value of 0 (Y channel of the CIE 1931 XYZ space), but non-zero chromaticity values (X and Z channels of the CIE 1931 XYZ space). This case is a numerical oddity that matches no physical reality (a pixel with no luminance should have no chromaticity too), will produce a division by zero in xyY and Yuv color spaces, and will create `NaN` RGB values as a result. This issue is **not** corrected inside _color calibration_ because it is a symptom of a bad input profiling and/or a bad black point level, and needs to be addressed manually either by adjusting the input color profile with the channel mixer or in the _exposure_ module _black level correction_.
```
matt-maguire, line 126: In fact, 0/0 is also undefined, so I don't think it matters whether the chromaticity values are zero or not.

aurelienpierre: It doesn't matter for numerical consistency, but non-zero chromaticity with zero luminance is a clue that leads to the diagnostic of bad color transform, because this usually happens in regions that should not be full black (I have seen that happen with blue stained-glass… backlit).

ralfbrown: For the particular case of highly-saturated blues, a major contributor is that at least half of the supported cameras have imaginary blue primaries in adobe_coeffs.c which result in camera->XYZ transform matrices that assign a negative weight to the B channel's luminance contribution.  So as the color becomes more blue, it eventually achieves zero or even negative overall luminance.

elstoc:
Thanks all

PR #287, opened 2021-06-24
Thumbnail cache default size is 512MB, not 256MB

victoryforce:
Thumbnail cache size was changed by this commit: darktable-org/darktable@c5f9e5dc8f


elstoc:
Thanks

PR #291, opened 2021-06-27
Fix grammar: targetted

victoryforce:
It should be no doubling when the preceding vowel is unstressed.

https://www.future-perfect.co.uk/grammar-tips/just-plain-wrong/is-it-targetted-or-targeted


elstoc:
Thanks @victoryforce

PR #292, opened 2021-06-29
Fix non-dictionary word: percentance

victoryforce:


elstoc:
Wow how did that sneak in there? Thanks @victoryforce

PR #296, opened 2021-07-09
Update official IRC channel to oftc.net

victoryforce:
I think it would be good to update this also in deployed docs on darktable.org in order to direct users to the correct channel, not abandoned one.


elstoc:
Agreed. Really we should just update every page that mentions freenode or github or the mailing list to point to the contact page and consolidate everything there.

elstoc:
So for this one, I think it's probably best to just say "raise an issue on github" and provide a link

victoryforce:
Agreed, we should probably emphasize more that the desired way of communication is github, even if other channels do exist.

PR #299, opened 2021-07-12
Remove duplicate words

victoryforce:
A handful of duplicate word removing. 

I accumulate commits for a while to avoid creating a large number of PRs, and group them by fix type.


elstoc:
Thanks

PR #300, opened 2021-07-19
Fix substitution variable name: it is SIDECAR_TXT, not SIDECAR_TEXT

victoryforce:
To check, the corresponding code is in the file src/commom/variables.c.


elstoc:
Looks good thanks

PR #303, opened 2021-07-19
Fix duplicate words

victoryforce:
More duplicates were noticed.


elstoc:
I've read the manual so many times. How come I've never spotted all this stuff! Thanks again

johnny-bit:
> I've read the manual so many times. How come I've never spotted all this stuff!

It's part of language processing to to skip repeated words. It was even used as a "gotcha" in couple reading tests using double negation.

elstoc:
@victoryforce has raised quite a few correction PRs (and much appreciated they are). It was more a comment on the sheer weight of things I've missed, not just these ones, but I take your point - repeated words are are easy to miss.

PR #307, opened 2021-07-21
Use american spelling

victoryforce:
As dtdocs use American spelling I change this word to be in conformity with the rest of the docs.

https://www.grammarly.com/blog/center-centre/


elstoc:
Ugh hate it ;) But you're right American spelling is the standard.

PR #309, opened 2021-07-23
There is no 'search folder' feature in collections, only 'search filmroll'

victoryforce:
You can verify this by looking at the source code of the collections module or by searching in darktable.po.


elstoc:
Indeed. Perhaps there should be. In the meantime the docs should reflect current functionality.

PR #311, opened 2021-07-24
The correct name for this kind of matrix is identity matrix

victoryforce:


elstoc:
Thanks. I did fix this on the color calibration module but forgot this one.

PR #313, opened 2021-07-25
Propose an example with EXIF date

Nilvus:
Propose an example to set import options with EXIF dates instead of current date as I see some times new users asking for that.
Following this discussion: https://github.com/darktable-org/darktable/pull/9558#issuecomment-882114312

### DIFF
--- a/content/module-reference/utility-modules/lighttable/import.md
+++ b/content/module-reference/utility-modules/lighttable/import.md
@@ -143,6 +143,8 @@ keep this window open
 
 Most of these options can also be set in [preferences > import](../../../../preferences-settings/import.md). See this section for more information about the available variables.
 
+For example, in all lines above, images are imported following current date. For those who want to use images dates (so EXIF ones), just add the `EXIF_` before each date part to have variables like `$(EXIF_YEAR)`, `$(EXIF_MONTH)`, `$(EXIF_DAY)`, etc.
+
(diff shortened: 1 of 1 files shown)


Nilvus:
@elstoc: your review is requested here (I can't set reviewers, labels or milestones here). Maybe my english will need improvement.

paperdigits:
I don't think this information fits well here, it seems random and we cover all the variables in another topic.

On July 25, 2021 9:13:54 AM PDT, Nicolas Auffray ***@***.***> wrote:
***@***.***: your review is requested here (I can't set reviewers, labels

Nilvus:
> I don't think this information fits well here, it seems random and we cover all the variables in another topic. 

It's not at all random as I see many users ask for such import. I'm so quite sure that many users would prefer import by EXIF dates instead of actual import date. That's why I suggested those as an example to help and show how to work with variables.  I think we could add that in the manual (or as a default in darktable import dialog window (but that would need a manual update too).

Maybe there's another place that fits well. I'm open to any suggestion.

paperdigits:
I agree that the default import date being today's date is a strange choice.

I think this information makes the most sense in the workflow topic about importing: https://darktable-org.github.io/dtdocs/overview/workflow/import-rate-tag/

Where you currently have it is all reference information, where it doesn't make sense.

PR #318, opened 2021-07-28
histogram: vertical waveform and scope appearance update

dtorop:
Describes the new "vertical" waveform and rgb parade modes (darktable-org/darktable#9170). Updates images due to visual legibility changes to these plots (darktable-org/darktable#9052) and add images for the vertical plots.

Some general wording updates, particularly as move toward using "scopes" as a generic name for the varieties of histogram/waveform/vectorscope.


elstoc:
So nice when I don't need to do any copy-editing after. Thanks @dtorop!

dtorop:
Thank you!

PR #334, opened 2021-08-12
Delete the phrase that documents the unreliable behavior of W and Ctrl+W shortcuts

victoryforce:
I want to suggest removing this phrase. I do not observe the mentioned unreliable behavior either in version 3.6 or in any post-3.6 git builds. Also, in any case, even if this bug still exists and can be reproduced, the right way is to fix the issue, not document it.


matt-maguire:
I don't think this is saying the feature is unreliable; it is saying that image needs to be properly selected in order for the W key to work. So, if a user tries to use the W key and it doesn't appear to be working, it is because they haven't selected the image properly. So, this remark is just giving advice to the user in order to preempt them filing useless bug reports.

elstoc:
Also can happen if the centre panel doesn't have focus or focus is currently given to an entry field. I think with NG input possibly only the latter affects shortcuts now though (or no image being selected).

matt-maguire:
Maybe we can explain that, so it doesn't make the interface seem flaky or random.

PR #337, opened 2021-08-28
fix darktable-generate-cache ranges of mips

matt-maguire:
On pixls someone pointed out the range of mips values in darktable-generate-cache is incorrect in the manual:
https://discuss.pixls.us/t/generate-full-resolution-images-in-cache/26599/4

This PR is to correcfly specify the valid range.


elstoc:
Thanks @matt-maguire

PR #341, opened 2021-09-05
lut-3D.md: add some info about types of LUTs

phweyland:


elstoc:
Thanks @phweyland. I'll merge and then do a copy-edit.

PR #351, opened 2021-09-22
Update Portuguese translation of configuration files

MarcusGama:


elstoc:
This certainly looks like a translation issue - I can see that the site title isn't translated on the website - though I'm not sure whether this is the right solution. @paperdigits can advise better here but my understanding is that all translations need to be done on [weblate](https://weblate.pixls.us/projects/darktable/dtdocs/). If there is an issue with the theme translations it would be better if we could resolve this on weblate and keep the translations there.

paperdigits:
This is the correct way to do it. Thanks!

elstoc:
Perhaps a message to the dev mailing list and an edit of the translations content page would be useful here because I'm not sure the process is clear. We have other translations that are complete on weblate but don't include this change.

PR #376, opened 2021-11-14
Change in the content.pot file.

lebmich:
New wording to clarify the concept of duplicate instance. I hope ...

### DIFF
--- a/content/module-reference/processing-modules/diffuse.md
+++ b/content/module-reference/processing-modules/diffuse.md
@@ -42,7 +42,7 @@ Since the process is physical, even its glitches may be used for creative purpos
 
 ---
 
-**Note:** This module is highly resource-intensive, as it is actually an anisotropic, multiscale, partial differential equation solver. The module's runtime increases with the number of iterations and OpenCL is therefore strongly recommended.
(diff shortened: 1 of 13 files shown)


elstoc:
Hmm I'm not convinced this is really any clearer. Also, you can't just modify the content.pot file, you have to change the original markdown files themselves.

lebmich:
In the English version of the user manual we have:

duplicate instance
    Create a new instance of the current module with all of its parameters **inherited from the parent module**.

I think the expression "inherited from the parent module" is not precise enough. What must be expressed is that the values of the parameters of the new instance are those of the parameters of the instance which is being duplicated.

paperdigits:
That is what the phrase "inherited from the parent" means though.

lebmich:
Maybe we could write "inherited from the parent **instance**"

lebmich:
or better : "inherited from **its** parent instance"

lebmich:
We could write:
duplicate instance
Create a new instance of the current module with its parameters settings inherited from its parent instance.

elstoc:
The rest of the section refers to "current instance" so we'll go with that

PR #378, opened 2021-11-15
Fix typos

amousset:
Using https://github.com/crate-ci/typos. Could also be added as a pull request test as it has low false positives.


elstoc:
> Could also be added as a pull request test

Could be, though I am a cowboy so I usually commit directly to the repo. Could you to raise a FR to suggest it.

Please remove the po and pot files from this PR as we have a separate process to update them independently of the markdown.

elstoc:
Thanks @amousset.

PR #381, opened 2021-11-16
Reflect renaming of 'max radius' parameter to 'radius span'

victoryforce:
This parameter was renamed in commit darktable-org/darktable@5cb2219


elstoc:
Thanks

PR #385, opened 2021-11-21
Fix an inaccuracy in the default export directory

victoryforce:
It is `darktable_exported`, not `darktable`.


elstoc:
Indeed. Thanks

PR #399, opened 2021-12-08
Replace the word with a more commonly used one

victoryforce:
This is not a typo fix or switching to American spelling, but a purely stylistic change.

Major English dictionaries lists spatial as the preferred word and spacial as the less commonly used alternative (M-W) or spelling variant (Collins).

I do not want to say that spacial is invalid word, but Google Ngram shows that the word 'spacial' is used many orders of magnitude less frequently:
https://books.google.com/ngrams/graph?content=spacial%2Cspatial&year_start=1800&year_end=2019&corpus=26&smoothing=3

### DIFF
--- a/content/module-reference/processing-modules/bloom.md
+++ b/content/module-reference/processing-modules/bloom.md
@@ -21,7 +21,7 @@ _**Note:** This module performs blurs in Lab color space, and is no longer recom
 # module controls
 
 size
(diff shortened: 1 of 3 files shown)


elstoc:
Agreed. Spatial is better. Will have to wait for the next version now though.

PR #427, opened 2022-02-13
Duplicate images on camera during tethering.

spaceChRiS:
Added information about how to keep images on the camera in parallel to the tethering download.


spaceChRiS:
Some background information can be found here: https://discuss.pixls.us/t/tethering-without-losing-the-images-on-my-camera/29366.

spaceChRiS:
Some activity for this PR to not get closed, I still think this information would be good to have in the manual.

spaceChRiS:
Some more activity … if this content is not wanted, pleas feel free to close it, but losing it just because time is running would be a pity.

elstoc:
Tethering is not something I use much so I'll take your word that this is a valid process! I'll merge and then do some copy-editing.

PR #429, opened 2022-02-14
We should not call the 'filmic rgb' module just 'filmic' for short

victoryforce:
We should not use the abbreviated name of the module instead of the exact one. Especially considering that such an abbreviated name belonged to another, now deprecated, module.


elstoc:
Seems reasonable to me.

PR #439, opened 2022-03-17
Fix repeated word

victoryforce:
I believe this is the last correction of this type (word repetition) for the current dtdocs text.


elstoc:
Cool. Thanks for all your work on this @victoryforce. Hard to believe we've made that mistake so often, but I guess it's easy to miss when reviewing.

victoryforce:
Yes, it's very easy to miss because our brain automatically corrects this when we read to understand the text. Therefore, to catch this type of typographical error, we need to focus solely on detecting duplication, without even trying to understand the text during this process.

PR #443, opened 2022-04-04
Grammar fix ('why' instead of 'that')

victoryforce:
'reason' is not used in the plural with a that-clause. Proof: https://dictionary.cambridge.org/grammar/british-grammar/reason


elstoc:
hmm I would be tempted to get rid of the word entirely: "there are three main reasons this could happen"

victoryforce:
OK, I modified PR.

elstoc:
Thanks

PR #444, opened 2022-04-06
The darktable database contains only information about the image, not the image itself

victoryforce:
The current text gives the impression that the image itself is stored in the library, so I changed it to avoid ambiguity in understanding.


victoryforce:
> Should probably be changed in the tooltip in darktable itself as well.

I think so too. In the tooltip and in the preference label as well. I'll prepare PR for this.

PR #445, opened 2022-04-06
A pixel cannot have an edge

victoryforce:
This is probably a mistake. A pixel is the smallest, "atomic" element of an image, it cannot have an edge.


elstoc:
I've no problem merging this but don't worry too much about making deprecated module documentation too perfect, since we no longer support these modules, and any change will just give another job to all our translators.

victoryforce:
OK. OTOH, being translator myself (so, yes, this will be job for me too) I'd say this is no big deal. Weblate makes modifying existing translation of changed (especially only slightly changed!) text a breeze, it takes, literally, a few seconds.

PR #446, opened 2022-04-06
Change the checkbox label in the metadata editor

victoryforce:
The metadata editor contains a checkbox with the opposite meaning: 'visible', not 'hidden'.

### DIFF
--- a/content/module-reference/utility-modules/shared/export.md
+++ b/content/module-reference/utility-modules/shared/export.md
@@ -119,7 +119,7 @@ exif data
 : Export the source image's Exif data.
 
 metadata
-: Export metadata defined in the [metadata editor](../shared/metadata-editor.md) module. Metadata fields tagged as _private_ or _hidden_ will not be exported.
+: Export metadata defined in the [metadata editor](../shared/metadata-editor.md) module. Metadata fields tagged as _private_ or not tagged as _visible_ will not be exported.
 
 geo tags
 : Export geo tags.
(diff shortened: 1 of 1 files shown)


elstoc:
hmm I'm not sure about this one. I get where you're coming from but "hidden" and "not visible" mean the same thing and I'm not entirely happy with how that pair of "nots" reads in this sentence ("or not tagged as _visible_ will not be exported").

I'm sure there's some less clunky sounding wording but I haven't thought of it yet...

victoryforce:
"hidden" and "not visible" surely mean the same thing.  But here we refer to a specific label in the GUI. So as long as "visible" is used in the program interface, we must refer to this exact name, otherwise the user will be confused by the discrepancy in the GUI and documentation.

I can swap two parts of this sentence so that 'nots' will be farther apart. So, intead of
`Metadata fields tagged as _private_ or not tagged as _visible_ will not be exported.`
it will be
`Metadata fields not tagged as _visible_ or tagged as _private_ will not be exported.`

I think that sounds better. Is it good enough for you or will you try to find an even better wording?

elstoc:
> Metadata fields not tagged as _visible_ or tagged as _private_ will not be exported.

That could be potentially be misread as "metadata fields not tagged as visible or not tagged as private...". How about if we reversed the logic and said something like "Only metadata fields that are tagged as _visible_ and are not tagged as _private_ will be exported."

victoryforce:
Reversed text sounds so much better! What amazes me is why I didn't come up with this. :)
So I modified the PR with you suggestion.

elstoc:
Thanks.

PR #459, opened 2022-05-19
Update variables.md

phweyland:
Add $(MSEC...


elstoc:
Does this relate to a darktable PR? If so would you mind including a link to the darktable PR just so I can keep track of what remains to do. Similarly with your other PR.

PR #462, opened 2022-05-19
tagging

phweyland:
- better key management #10784
- improved suggestions list #11030
- dragging, expand bypass #10874


elstoc:
Thanks @phweyland. I've pushed a copy-edit. There were some sections that weren't entirely clear so I've edited based on what I think they were intended to say. I'm still not sure I understand how the "level of confidence" preference works though. Feels like a 100% match should show a single tag rather than none or am I missing a key part of the description?

phweyland:
> Feels like a 100% match should show a single tag rather than none or am I missing a key part of the description?

A tag with 0% of confidence has no chance to be correct. A tag with 100% of confidence should be surely a good one. If you select 0%, you get all the tags, even the non matching ones, 100% for the list of tags which are fully matching (more selective). 
As I needed a simple way to disable this algorithm I used the 100% value for that (like an unreachable level of confidence), knowing that 99% of confidence is already almost fully matching.

I see I have make a mistake for the key Enter and Shift+Enter. The related view is the dictionary view, NOT the attached one.

elstoc:
I've pushed another edit. Does it look better now?

phweyland:
Yes. Looks good. Thanks.

elstoc:
Thanks

Issue #171, opened 2020-12-04
Add Windows/macOS file locations to the manual


Neverest:
In page 15, in the pdf, or in the basic workflow ->editing an image: scene-referred workflow, in the vibrance part it says this:

"vibrance

Tends to darken colors. Consider using using color zones with a saturation parametric mask to give more control."

The word using is repeated there.

Also, I noticed that in the manual it only mentions the location of the local copies, profiles etc of Linux systems. It would be nice if the Windows and macOS file locations where there too.

elstoc:
> The word using is repeated there

Thanks. I'll fix this shortly

> only mentions the location of the local copies, profiles etc of Linux systems

It generally uses variables like `$HOME` to reference the starting portion of the filename. You should replace this $HOME variable with the required path for your OS. Some of these paths are even different between Linux distributions and, certainly for me, I have multiple different installations with different paths defined. Could you point out a specific instance that you think needs to be amended please and why the above formulation doesn't work for other operating systems.

Neverest:
Ok should I start another issue then?

elstoc:
Probably no need now but how about you change the title.

Neverest:
> Could you point out a specific instance that you think needs to be amended please and why the above formulation doesn't work for other operating systems.

Well, in Windows the darktablerc, keyboardrc, library.db and data.db are located in `C:/$USER/AppData/Local/darktable` which by default you first need to go to your File Explorer -> View -> check Hidden Items to be able to see it.

The local copies, log and mipmap files are in `C:\$USER\AppData\Local\Microsoft\Windows\INetCache\darktable` and to be able to see it you need to do the above and go to File Explorer Options -> View tab -> uncheck "Hide protected operating system files".

I mention them because there is a chance the user doesn't know how to make those folders visible, especially the INetCache one. Just yesterday I recommended this program to a friend, that didn't know where the AppData was located, so the less friction from non computer savvy people the better.

For where it would be convenient, in the local copies sections, and the first time the darktablerc file is mentioned. Sorry I can't think of any more areas, since I'm new at Darktable and photography in general.

 It could be a small section called "Configuration File Locations" with a list of all the locations and a brief description of what that file has sort of like[ kdenlive does it](https://userbase.kde.org/Kdenlive/Manual/Installation) but adding the other OS.

paperdigits:
This is covered in the FAQ on the website, and the path for WIndows is `C:\Users\[username]\AppData\Local\darktable\`

[...]

PR #467, opened 2022-05-21
Update CSS part

### LINKED ISSUE #414: The CSS/Themes section needs to be updated.
Currently this section is incomplete and does not fully reflect the best (only?) way how a user can use/implement their own theme/style.

This issue is a companion to this darktable issue: [ Theme: css file placed in ... #10914 ](https://github.com/darktable-org/darktable/issues/10914) and [this](https://discuss.pixls.us/t/my-personal-darktable-theme/28942) Pixls topic.

Nilvus:
This fix #414 

@jade-nl, @elstoc and @difrkaguilar: this improve CSS part as requested on #414 and follow so links provided by @jade-nl in that issue.

Those links are old now, so even if I read some parts to work that, it's possible that I miss something. So your feedback are more than welcomed to be sure this will correctly improve that part.

And of course, as my english is not my first language, I'm quite sure english is not the best one!

### DIFF
--- a/content/preferences-settings/general.md
+++ b/content/preferences-settings/general.md
@@ -27,7 +27,7 @@ GUI controls and text DPI
 
 # CSS Theme Modifications
 
-In addition to selecting a pre-built theme you can also apply additional CSS customisations of your own to tweak the look-and-feel of darktable. darktable provides a textbox for this purpose in general tab of main preferences. No need to set import lines of darktable CSS themes. You just need to choose which default darktable theme you want to start from (in theme menu just above) and then just copy/paste tweaks you find or create your own.
+In addition to selecting a pre-built theme you can also apply additional CSS customisations of your own to tweak the look-and-feel of darktable. A text box is provided for this purpose. No need to set import lines of darktable CSS themes. You just need to choose which default darktable theme you want to start from (in theme menu just above) and then just copy/paste tweaks you find or create your own.
 
 When you have finished entering your CSS tweaks, click on the 'save CSS and apply' button. This will save your CSS to a file (`$HOME/.config/darktable/user.css` on Mac and Linux ; `C:\%LOCALAPPDATA%\darktable\user.css` on Windows) and immediately apply it to the current darktable session.
 
@@ -37,6 +37,7 @@ If your changes cause any issues, you can uncheck the "modify selected theme wit
 
 If you want to understand how darktable themes are constructed you are advised to look at the file `$DARKTABLE/share/darktable/themes/darktable.css` for more information (where `$DARKTABLE` is darktable's installation directory). You can also use the Gtk Inspector tool to analyse the existing layout.
 
-Now darktable CSS use classes for main parts and then ids for specific ones, like it is recommended by W3C specifications. Main darktable classes are now specified in darktable.css header. Other ones could be found inside file content. To optimize CSS, set on text box only what you want to change. No need to copy parts already set in darktable.css file (or other theme you select). It's always better to set tweaks that will impact all items (for example: buttons) then set specific tweaks if you want to just set a precise tweak for specific button.
+Now we have darktable classes first used for main items (buttons, or items with transparent background for example), then CSS ids are set for specific items (for example modules tabs that are Gtk buttons but with different behavior than classic buttons). Those classes are now specified in darktable.css header. Other ones could be found inside CSS code.
+To optimize CSS, set on text box only what you want to change. No need to copy parts already set in darktable.css file (or other theme you select). It's always better to set tweaks that will impact all items (for example: buttons) then set specific tweaks if you want to just set a precise tweak for specific button.
 
-*Note that actual darktable themes are without colors (except some needed ones like color labels). This is done on purpose. The goal of darktable UI is to let most place to the images and avoid any distractions.*
+*Note that actual darktable themes only use shades of gray (except some needed ones like color labels). This is done on purpose. Colourful interface might distract the eye from the image.*
--- a/content/preferences-settings/general.md
+++ b/content/preferences-settings/general.md
@@ -27,7 +27,7 @@ GUI controls and text DPI
 
 # CSS Theme Modifications
 
-In addition to selecting a pre-built theme you can also apply additional CSS customisations of your own to tweak the look-and-feel of darktable. You could create your own CSS file but this will need to choose which theme to import and set it's path hardcoded in your CSS file depending on your system also. This is not the best way. darktable provide a better and simpler way: a text box is provided for this purpose in general tab of main preferences. No need to set import lines of darktable CSS themes. You just need to choose which default darktable theme you want to start from (in theme menu just above) and then you just have to copy/paste tweaks you find or create your own.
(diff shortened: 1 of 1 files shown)


jade-nl:
Just had a look at the proposed changes; This looks a lot better then before.

I do see some things that could probably be phrased a little better, but, like you, I'm not a native English speaker so maybe @elstoc can fix/change those.

I'm not sure about the part that describes that you could create your own CSS but shouldn't, this part:

> You could create your own CSS file but this will need to choose which theme to import and set it's path hardcoded in your CSS file depending on your system also. This is not the best way. darktable provide a better and simpler way. darktable provide a better and simpler way: a text box is provided for this purpose [...]

Keep it terse and don't mention that part. Maybe something like this instead of the above quoted part:

> darktable provides a textbox for this purpose [...]

Just my 2c, as always.

Nilvus:
I hesitate to add that part. I add it because I read some comment about that on links you posted. But was also wanted to not precise that. I here decide to follow comments I read but I agree with your answer so last commit remove that following your proposal.

I also update others parts. That should be, I hope, better english.

content/preferences-settings/general.md
```diff
@@ -27,12 +27,16 @@ GUI controls and text DPI
 
 # CSS Theme Modifications
 
-In addition to selecting a pre-built theme you can also apply additional CSS customisations of your own to tweak the look-and-feel of darktable. A text box is provided for this purpose.
+In addition to selecting a pre-built theme you can also apply additional CSS customisations of your own to tweak the look-and-feel of darktable. darktable provides a textbox for this purpose in general tab of main preferences. No need to set import lines of darktable CSS themes. You just need to choose which default darktable theme you want to start from (in theme menu just above) and then just copy/paste tweaks you find or create your own.
```
elstoc, line 30: You don't need to mention the "general tab of main preferences" since this entire section of the user manual is _about_ the general tab of preferences.

Also, we shouldn't start sentences with lowercase letters. In order to not annoy the "darktable always has a lower case d" people, I therefore avoid at all costs putting "darktable" at the start of a sentence (hence the previous wording).

content/preferences-settings/general.md
```diff
@@ -27,12 +27,16 @@ GUI controls and text DPI
...
 
-When you have finished entering your CSS tweaks, click on the 'save CSS and apply' button. This will save your CSS to a file (`$HOME/.config/darktable/user.css`) and immediately apply it to the current darktable session.
+When you have finished entering your CSS tweaks, click on the 'save CSS and apply' button. This will save your CSS to a file (`$HOME/.config/darktable/user.css` on Mac and Linux ; `C:\%LOCALAPPDATA%\darktable\user.css` on Windows) and immediately apply it to the current darktable session.
```
elstoc, line 32: I don't use windows much but is this location always on the C drive? Or is there a more generic thing (like $HOME) that we can use. It's worth noting that this will be the only place we reference windows file locations so, for consistency it might be better to stay Linux/Mac-only here and update them all at once as a separate change.

content/preferences-settings/general.md
```diff
@@ -27,12 +27,16 @@ GUI controls and text DPI
...
 
-If you have any issues with darktable please retry with this option unchecked to be certain that they were not caused by any of your alterations.
+*If you have any issues with darktable please retry with this option unchecked to be certain that they were not caused by any of your alterations.*
 
 If you want to understand how darktable themes are constructed you are advised to look at the file `$DARKTABLE/share/darktable/themes/darktable.css` for more information (where `$DARKTABLE` is darktable's installation directory). You can also use the Gtk Inspector tool to analyse the existing layout.
+
+Now darktable CSS use classes for main parts and then ids for specific ones, like it is recommended by W3C specifications. Main darktable classes are now specified in darktable.css header. Other ones could be found inside file content. To optimize CSS, set on text box only what you want to change. No need to copy parts already set in darktable.css file (or other theme you select). It's always better to set tweaks that will impact all items (for example: buttons) then set specific tweaks if you want to just set a precise tweak for specific button.
```
elstoc, line 40: I'm not 100% sure what you're trying to get across here...

- what do you mean by "main parts" and "specific ones"? I'm sure there's a better way of putting it but perhaps it needs to be more clear. No need to mention W3C specs IMO.
- what do you mean by "inside file content"?
- I'm not sure what you're trying to get across with "It's always better to set tweaks that will impact all items (for example: buttons) then set specific tweaks if you want to just set a precise tweak for specific button."

content/preferences-settings/general.md
```diff
@@ -27,12 +27,16 @@ GUI controls and text DPI
...
+
+*Note that actual darktable themes are without colors (except some needed ones like color labels). This is done on purpose. The goal of darktable UI is to let most place to the images and avoid any distractions.*
```
elstoc, line 42: By "without colors" I assume you mean to say something like "darktable themes only use shades of gray". I think most people think of gray as a colour.

When you say "The goal of darktable UI is to let most place to the images and avoid any distractions." you are presumably talking about the fact that a colourful interface might distract the eye from the image?

elstoc (commented):
No need to get too hung up on the wording for now, since I'll probably copy-edit it into better sounding English after merging. But there are a few places where I'm not sure what point you're trying to make, so if you could add some clarification that would be much appreciated.

content/preferences-settings/general.md
elstoc: Or perhaps even provide a single section about file locations and just link to it from everywhere.

Nilvus: By default, it should be only on C: drive and as far as I know, there's no more generic thing that what I write here. I follow some instructions provided by Windows user.
About a specific section, probably better but this section should be set first. This text could be updated later when that section will be done. It's a request of this issue: #171

content/preferences-settings/general.md
Nilvus: I will update and simplify that part to follow your comments. Hope it will be better and clearer. For last point, It's just about CSS recommendations about what we never do before and I updated on CSS code. If you want to change the UI: for example, we set first CSS settings that will apply to all buttons. And after that only set some needed tweaks for specific buttons. So first CSS apply to a lot of items then set CSS settings that apply to only some. I hope you will find with that explanation the way to better write that in english.

content/preferences-settings/general.md
Nilvus: Yes for both. And indeed shades of grey is better.

Nilvus:
Last commit should be better. @elstoc: I hope this (and my comments) would help you to improve my text with better english.

jade-nl:
Looks a lot better now. Just 2 things I noticed:

1. darktable uses America-English so _Colourful_ should be _Colorful_
2. This sentence doesn't seem to be correct/clear: 

> Now we have darktable classes first used for main items [...]

Not sure how to improve it, but the previous version, even though it is somewhat bloated, is clearer to me. I'm sure that @elstoc can knead this into something that is clear to all of us.

content/preferences-settings/general.md
elstoc: But you're talking about how the theme CSS works, rather than how a user should tweak it?

Nilvus: it's the same. a user need to take care of that to tweak css

difrkaguilar:
> And of course, as my english is not my first language, I'm quite sure english is not the best one!

English is not my native language either. But I'm always trying to do my best... and using DeepL too jjjjj

elstoc:
Rather than merge this and then push copy-edits afterwards, I have created a separate PR (#471) with some additional wording. I hope I have included the key points from this PR but please feel free to review.

PR #473, opened 2022-05-26
Fix the slider description in vignetting module

victoryforce:
The current description of the action of this slider is the exact opposite of its behavior in reality


elstoc:
Thanks!

PR #476, opened 2022-05-30
print view: describe new overlays

dtorop:
Describe the overlays in https://github.com/darktable-org/darktable/pull/11446.


elstoc:
Thanks @dtorop!

dtorop:
Welcome! At some point could probably use an image,but it should still be clear enough...

PR #480, opened 2022-06-04
(WIP) document v4.0 openCL changes

### LINKED ISSUE #449: Re-write all performance related settings
@elstoc (and all people interested in the docs) this is somewhat demanding, first some general ideas

1. Should we keep the 32bit section at all?
2. I would write a section about "tiling - what is the relevance of it" both for cpu and gpu
3. A general introduction about the idea of "resourcelevel". This would include
   a) scenarios in which to use what level
   b) hints for the "nerds" about what the fractions mean. I personally think this part should be in the section we now use for memory and stuff like that. Maybe even call that "Nerds tuning"?
4. The "opencl tuning mode" should
   a) describe the checking for available memory
   b) mention the memory transfer mode used for tiling. That also have a hint that switching that on might be good on some systems, not so good on others.
   c) mention the new per_device conf key that could be tuned
   d) the internals of the device conf key should also go to the "nerds section". Some descriptions of those global OpenCl settings seem to be at least pretty outdated and maybe even wrong.

What approach would you prefer? Would you want me to write all docs as plain ascii and you modify the git repo? Or some pr's - (as i don't generate the docs myself and rely on the auto generated they might have errors you need to fix later) ?

elstoc:
Resolves #449


content/preferences-settings/processing.md
```diff
@@ -80,3 +81,10 @@ OpenCL scheduling profile
 : - _default_: the GPU processes the center view pixelpipe; the CPU processes the preview pipe; 
 : - _multiple GPUs_: both pixelpipes are processed in parallel on two different GPUs; 
 : - _very fast GPU_: both pixelpipes are processed sequentially on the GPU. 
+
+tune OpenCL performance
+: Defines how darktable will attempt to tune OpenCL performance for your system. The following options are provided (default _nothing_):
+: - _nothing_: do not attempt to tune OpenCL performance.
+: - _memory size_: on the first run of a pixelpipe, darktable determines how much memory is available and assumes that the same amount will continue to be available for the remainder of your darktable session. This will usually be safe if you are not using any other applications that need a reasonable amount of your GPU's memory. Otherwise you might be better to disable this option, as it could lead to out-of-memory errors.
```
jenshannoschwalm, line 88: I would say
1. "This will usually be safe if you are not starting other applications that need a reasonable amount of your GPU's memory while running darktable"
2. " ... as it could lead to out-of-memory errors resulting in a performance drop because of falling back to cpu code."

Also maybe some additional info like
"You may switch this off at any time and on again to re-tune"

jenshannoschwalm:
Otherwise i see no "technical errors"

jenshannoschwalm:
Looks good for me when viewing the diffs ...

elstoc:
> Looks good for me when viewing the diffs ...

Excellent. Thanks for reviewing. I'll probably do another quick copy-edit and then merge it later today.

elstoc:
@jenshannoschwalm: one question about the version number. I assume that this is an internal thing that will be incremented when the parameters change in a new version of darktable? What will happen with previously-overridden settings when this happens? Say I have tweaked version 4 and you create a version 5 - would my tweaks be carried over or would I have to re-tweak? If the latter, how would I know the version has changed?

jenshannoschwalm:
Oh, my comment from mobile seems to be gone.

Yes, internal version but kept in darktablerc as before.
NEW: whenever the version bumbs there can be a note shown requesting a confirm by the user like "understood"? If answered yes that note won't be shown again, otherwise whenever dt starts.

So far whenever there was a dt version jump as from 3.8->4.0 we made sure conf data are kept. This will be so later too. (I don't think the code is so messed it can't be done :-)

jenshannoschwalm:
Have looked through your late mods, all changes are good for me regarding technical aspects.

Style wise, thank you is by far not enough!

elstoc:
Thank you for your kind words :)

PR #481, opened 2022-06-11
document additional midi config options

elstoc:


elstoc:
@dterrahe: please can you review

dterrahe:
The explanation i gave was intended to clarify it to _you_ so you could rephrase it in the best way to target a larger audience. As such I wasn't aiming for brevity or paying attention to any style or grammar requirements. I cannot really opine if the format is suitable for wider dissemination unchanged as you are proposing doing here; I'm simply too deep into this stuff to have any perspective. Sorry.

elstoc:
I don't / won't use midi so the best I can do is repeat what you've written - I've amended the language a little but I can't test it or anything so if you're happy that it's _technically_ correct the best we can do is merge as-is and see if we get any feedback.

dterrahe:
Fine with me, of course.

> see if we get any feedback

I think I could give you a rough estimate of the number of comments to expect...

The individual topics for each device could be extended with their respective config blurbs, where they are "necessary" (BeatStep and Loupedeck), replacing the bit about slowly turning left 5x. But it would be nicer if individuals actually owning those devices step up and update the docs. After all, I cannot test anything I don't own either and I also cannot verify if there is enough handholding in the text for a newbie.

Maybe adding something to the top of the topic to invite contributions and feedback could help (ha!). Something like:

> If you have some success in getting a midi device up and running that hasn't been mentioned below, it would really be appreciated if you would provide feedback and possibly help for others if some special steps are required. You could do this either here, by submitting a PR to add a subtopic to the text or by filing an issue with the information. Or you could respond to any of the (closed) midi or inputng requests or PRs in the main darktable repository if you need additional help or guidance yourself.

elstoc:
> I think I could give you a rough estimate of the number of comments to expect...

Indeed. Would be interesting to know how many people are using midi and with what devices.

> Maybe adding something to the top of the topic to invite contributions and feedback could help (ha!).

Couldn't hurt. I'll add something.

elstoc:
Let's merge...

PR #488, opened 2022-06-15
Adding new CL device config key

jenshannoschwalm:
Ok,i'll give it try. Feel free to modify as you think ...


elstoc:
Not sure I entirely understood what you were going for in https://github.com/darktable-org/darktable/pull/12058 but will that affect this change?

jenshannoschwalm:
Yes, the default changed to 400.
Will update the pr

jenshannoschwalm:
It should be like this
*******************************************
There is a second device specific configuration key taking care of parameters that are related to the device hardware **plus** the device id, the usual key name `cldevice_version_canonicalname` is followed by `_idX` with X being the device id.

For darktable 4.0 there is only one parameter yet defined
a. forced headroom (only valid if tuning for memory is switched on)
: 0 test for currently available memory on this OpenCL device, this avoids manual tunings or checks. The code might not work on your device / driver combination and might lead to instability or even crashes.
: headroom (in MB) defines how much of this cards memory will not be used by darktable (default = 400).

It helps to get the best performance if you are sure that no apps or the OS make use of a specific device so darktable can take all of it. Likely this can be done if you have more than one OpenCL device available on your system, in this case set to 1 for the otherwise unused devices.

The default 400 is likely ok on most systems, if you run into performance drops due to cpu fallbacks you should disable "tune for memory" first. If yiu are sure your desktop always takes more than 400 increase this to 600.


*******************************************

jenshannoschwalm:
Looks good to me, one minor thing yet, you may switch memory tuning off and on again to force another check at runtime.

elstoc:
I added that to the new section in the perf tuning but not to the preferences section since the default behaviour no longer does any checking.

elstoc:
Ok I'm going to merge this now. Let me know if you spot any further changes needed.

PR #497, opened 2022-06-22
Correct one typo and one factual error

victoryforce:
This commit fixes one typo ("be" instead of "by") and one factual error: the opposite value, not the specified one, is correct.

Also "set value to 1" is not an accurate wording as here "1" is the value, so it's better to write "set parameter" (or field, or setting as other options).


elstoc:
Thanks @victoryforce. Since we're already editing this file in another PR I've added these changes there and I'll close this one. See #488.

PR #502, opened 2022-06-27
Typo ans style fixes

victoryforce:
The name of the option is 'graphic', not 'graphics'.

Also, there is no point in aligning the colons. They will only be aligned when viewed in a fixed-width font environment, this will not be the case in a browser window.


elstoc:
You missed the typo in your commit name ;-)

victoryforce:
> You missed the typo in your commit name ;-)

Yeah, I'm too sleepy now. :-)

PR #503, opened 2022-07-01
Correct the filter name to be more accurate

victoryforce:
While it's OK to call the filter "text search" for clarity while the label in the UI says just "search", it would be better to not change singular to plural when referring to the "color label" filter.


elstoc:
Only just made it in time. Will be string-freezing very soon.

PR #516, opened 2022-08-01
separate file extensions among raw, dynamic range and ordinary types

paolobenve:
The file types are now separated among raw types, dynamic range types, and ordinary types.

Added MRW into raw types, as in `/src/common/imageio.c` around line 80


elstoc:
The documentation used to split like this (between HDR and LDR) and I removed it because I didn't think it really added anything. Raw and non-raw is enough IMO.

PR #525, opened 2022-09-30
Give sample information pattern for OpenCL warning

dterrahe:


dterrahe:
It might also be useful to add this tip to the opencl section, but I'm not sure from an educational perspective what the best location would be.

elstoc:
Let's put this in a section of its own, then we can add a link to the OpenCL section. I'll merge this and edit.

PR #534, opened 2022-12-01
Update process.md

gi-man:
Remove the "better" approach of using filmic RBG highlight reconstruction and turning off the Highlight Reconstruction module. This is just not really accurate and more so with 4.2 new algos in HL module.


elstoc:
I agree that this section will probably need some editing but I'm reluctant to remove mention of the filmic reconstruction entirely. I haven't really used any of the new HL reconstruction stuff yet and am hoping that @jenshannoschwalm will be able to submit a first draft of the module reference documentation with some recommendations on what to use the different methods for, and perhaps some advice for this section.

jenshannoschwalm:
I have been working quite hard on code still...

Some manual article mostly done by @Iain is ready for your review..

elstoc:
> Some manual article mostly done by @iain is ready for your review..

Is it ready somewhere I can see it?

gi-man:
https://docs.google.com/document/d/1iMEGGL-zD_8v4NlSuOQgh1ysWWNXouqcP3b7K_1OY-k/edit?usp=sharing

jenshannoschwalm:
Will post it tonight

jenshannoschwalm:
[Darktable new highlight reconstruction method user documentation.odt](https://github.com/darktable-org/dtdocs/files/10142965/Darktable.new.highlight.reconstruction.method.user.documentation.odt)

Here it is

jenshannoschwalm:
I didn't look into this part of the manual yet, @gi-man i fully agree it's totally outdated and wrong for 4.2.

1. The phrase `darktable provides some tools handle blown highlights, in some cases even reconstructing lost colors or structure.` probably would be better as "darktable provides some tools handling blown highlights reconstructing lost colors and structure."
2. This `The [highlight reconstruction](https://darktable-org.github.io/dtdocs/en/module-reference/processing-modules/highlight-reconstruction/) module provides a very basic attempt to manage blown highlights. By default, if any color channel is clipped, this module will clip the other channels, making the pixel white – this prevents darktable from showing unnatural colors in the highlights. However it can be altered to attempt to reconstruct the lightness or color using any un-clipped channels. While this can re-introduce some detail, beware of the unrealistic colors that can result.` is also no correct any more. It should be like "The [highlight reconstruction](https://darktable-org.github.io/dtdocs/en/module-reference/processing-modules/highlight-reconstruction/) module by default provides an inpainting algorithm making use of channels not affected by being blown out recovering color and detail in most cases."
3. `A better alternative is to disable highlight reconstruction and instead use the [reconstruct tab on the filmic rgb module](https://darktable-org.github.io/dtdocs/en/module-reference/processing-modules/filmic-rgb/#reconstruct).` That is completely outdated again. I see two options here a) remove the complete paragraph - for me this is simply misleading now or if you want to keep a reference to that " An alternative would be to disable highlight reconstruction and instead use the [reconstruct tab on the filmic rgb module]

elstoc:
> An alternative would be to disable highlight reconstruction and instead use the [reconstruct tab on the filmic rgb module]

My (albeit limited) understanding was that these two modules were trying to do different things (that HL reconstruct could do some work early in the pipe and then filmic builds upon it later?). I'm still reluctant to lose the filmic paragraph but I'm not sure how best to separate the use cases for the two options (or if my understanding is outdated now).

jenshannoschwalm:
I think thee new modes just work better in 99% of images.. A few images still can take advantage, artificial multiple light sources and alike

elstoc:
> Since 4.2, filmic HL disabled by default since the HL module handles the clipping much better

Filmic HL reconstruction has always been disabled by default (though in the past this was done by setting the threshold to a high value). It was originally done so as not to degrade performance.

PR #535, opened 2022-12-03
Add sigmoid module reference

jandren:
Proposal dtdoc entry for the new sigmoid module.

https://github.com/darktable-org/darktable/pull/12642

I guess further contributions are needed in other parts of the docs but figured this was a good start!


jenshannoschwalm:
A nice "don't use with ..." section, hopefully read by users :-)

elstoc:
Just some initial comments from a brief read-through:

- The first sentence (easy and robust) should be removed - in my opinion it's subjective and doesn't add anything
- Some of the usage notes are more general notes about scene-referred and display transforms, so might work better in another section (perhaps the overview section or the [module order](https://docs.darktable.org/usermanual/4.0/en/darkroom/pixelpipe/the-pixelpipe-and-module-order/) section). I'm open for suggestions.
- As a general principle we avoid images in dtdocs unless they are absolutely necessary. Although they are a nice illustration, I'm not convinced we need them here and the descriptions are sufficient (or could be made so).

jandren:
Thanks for the comments!

@jenshannoschwalm I only have two do nots: only use one display transform and acknowledge the difference between scene referred and display referred pipeline position. Do you feel like both are covered well enough?

@elstoc Easy to remove first line, should I try to replace it with something else or just leave out a soft intro all together?

I will see if I can shorten the text about module order and add the link as a read more reference.

About the images, I did not that there is a general absence of images in the manual. I was actually first thinking about generating a matrix with example images but opted for the more focus approach with gradients. Would you like to remove them entirely, reduce the number of them, reduce the size in the repo? What are the constrains we work with here?

jenshannoschwalm:
About only one display transform. Yes I think explained well, maybe some **bold** don't do 😀 but people will still do it and post positive results on forum. We just have to emphasise again and again.

I always sort of disliked the "scene referred" term, don't know anything better. And now all dt people use that term so we stay with it.

Some other known guy used phrases like whiter than white, supernova and alike. Maybe that could help users to understand the unbound data and why we want that but have to do a transform.

I like the pics BTW.

elstoc:
> Easy to remove first line, should I try to replace it with something else or just leave out a soft intro all together?

Ideally it should be a brief introduction about what the module _does_ and no more. Try to mimic the sort of thing we have for other modules. I will copy-edit later so as long as it's _correct_ it doesn't necessarily have to be well worded.

> Would you like to remove them entirely, reduce the number of them, reduce the size in the repo? What are the constrains we work with here?

Partly it's about future maintainability (there are _very_ few contributors to this repo and it's always much easier to write text than generate images), and also about consistency (if we have lots of images for one module and none for another, without a good reason, it looks odd. If someone adds a new option later but is unwilling to regenerate the images, we'll end up removing them all anyway. For those reasons (and probably others I've forgotten) we've generally adopted a principle of "don't use images when words will do". In my opinion these images could be replaced with words without losing anything so I would prefer to do that unless you can make a really strong argument for including them (why words are _not_ sufficient).

paperdigits:
I don't find the image necessary in the reference for a module itself.

If we wanted to expand the scene referred section or something like that, a comparison might be useful, but is probably a maintenance burden as things will continue to shift over time.

jandren:
Thanks that are some good insights! Stay tuned for a update to the PR

jandren:
@elstoc removed images and changed the first line. I would suggest we add the link once that page exists. I.e. another PR

elstoc:
Since we're coming up on the release very soon and I don't have a great deal of free time at the moment, I'm going to merge this now, just so that some documentation is available for users. I'll come back and do some edits when I get the time.

PR #537, opened 2022-12-14
Add a descripton of the properties section in the mask manager module

paperdigits:
per https://github.com/darktable-org/darktable/pull/12561

The screenshot might also warrant an update.


elstoc:
Ok let's merge now and we can update the image later

PR #541, opened 2022-12-23
highlight-reconstruction: document darktable 4.2 changes

elstoc:


jenshannoschwalm:
Welldone - thanks a lot. I think it's just fine and nothing seems to be missing for me.

elstoc:
Cool, thanks for reviewing

jenshannoschwalm:
Just had a look at the manual as we have it now, some comments we might want to discuss.

1. The "reconstruct color" mode is available any only for old edits, now deprecated because of too many problems and inferiour quality.
2. It might be worth to mention that both "inpaint opposed" and "segmentation based" are
  a) designed to be working fine in scene referred mode and
  b) don't require (in fact they don't want to do so) any smoothing like in "laplacian"
3. **This part** `Note: When using the highlight reconstruction included with the [filmic rgb](https://darktable-org.github.io/dtdocs/en/module-reference/processing-modules/filmic-rgb/) module it may be better to avoid using this module in clip highlights mode (so that filmic rgb has more information to work with).` is not a good advice any more i think. "opposed" mode offers even more data for the filmic algorithm.
4. **This part** `All in all, you are advised to use the guided laplacians highlight reconstruction mode to: smooth the boundaries of clipped areas, recover spotlights and clipped areas of diameter below approximately 256px (on the full-resolution RAW), remove chromatic aberrations, which can occur during [demosaicing](https://darktable-org.github.io/dtdocs/en/module-reference/processing-modules/demosaic/) (the next module in the pipeline) at the boundary between clipped and valid regions.` is again somewhat outdated (I might have a personal bias here though) For me "good for small specular highlights" especially if different light sources are involved.

elstoc:
> a) designed to be working fine in scene referred mode and

I'm not sure what this means. HL reconstruction has always been in the scene-referred part of the pipeline and can't be moved.

jenshannoschwalm:
> I'm not sure what this means. HL reconstruction has always been in the scene-referred part of the pipeline and can't be moved.

Well, technically this works in rawspace so both "scene/display referred approaches" are involved. I was hinting to: both algos as the laplacian keep the white point - this makes them more suitable - like the laplacian.The other modes adapt to the existing one. But this all might be too technical.

elstoc:
Perhaps. I'm trying to walk a line between technical enough to be useful but not so technical that we need to explain the explanations.

jenshannoschwalm:
:-) It's your say for me.

PR #542, opened 2022-12-25
Document balanced tililing

### LINKED ISSUE #539: Documenting the "balanced OpenCL vs CPU tiling" option
### Rationale
On many machines we find pretty fast multicore cpus and a lot of system ram but a significantly smaller gpu memory with somewhat resrticted number crunching power. There is no problem using OpenCL code in modules as long as that doesn't lead to excessive tiling with large overlapping areas. Such a situation was rare, now we have as examples D&S or laplacian highlights which might behave bad and it would have been better to use cpu code instead.
************

In DT 4.2 we have the CL device specific conf key expanded by an **advantage** value, it is set to zero by default.

### How does this work
While processing the pixelpipe we estimate overall workloads for both OpenCL and CPU (this is **not** static but depends on parameters set in the module thus defining tile size, number of tiles, tile overlap ...). In most cases we would prefer the OpenCL code to be executed even if that would mean tiling as the graphics cards is mostly much faster (often 10 times faster if it is a dedicated card).

If the ratio of estimated workloads (CPU vs GPU) is larger than the **advantage** hint, dt will use the CPU for processing that specific module this time.

Unfortunately there is no way to use some sort of benchmark so far, you have to find out the GPU advantage for your system yourself. 

### How to set up the advantage hint?

1. Measure performance (-d perf) in laplacian highlights after setting "diameter of reconstruction" to something high.
2. While doing so make sure (by -d tiling) that the cl code doesn't tile.
3. Check for execution times with opencl on and off.
4. The advantage option can now be set to approx (cpu-time / gpu-time)

jenshannoschwalm:
- first try to document

Resolves #539

### DIFF
--- a/content/special-topics/mem-performance.md
+++ b/content/special-topics/mem-performance.md
@@ -92,13 +92,11 @@ If you want to make maximal use of your GPU memory for OpenCL, you have three op
 
 ## balanced OpenCL versus CPU tiling
 
-On many machines we find pretty fast multicore cpus with a lot of system ram but a significantly smaller gpu memory with somewhat restricted number crunching power (like typical integrated graphics). There is no problem using OpenCL code in modules as long as that doesn't lead to excessive tiling with large overlapping areas because of large requirements.
-In such cases it might be better to run untiled cpu code instead of heavy tiling on your gpu.
+In most cases, running a processing module on a high-powered GPU (the OpenCL codepath) is significantly faster than running the same module using the CPU codepath. However, many users have fast multi-core CPUs with a large amount of system RAM, but a GPU with significantly lower capabilities (typically, integrated graphics with small amounts of dedicated memory). Use of OpenCL code in these cases can lead to excessive tiling, and it is often better to run a module without tiling using the CPU codepath than to attempt to use OpenCL with heavy tiling.
 
-While processing the pipeline we estimate overall workloads for both OpenCL and CPU (this is not static but depends on parameters set in the module thus defining the tile size, number of tiles and tile overlap.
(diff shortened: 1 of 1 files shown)


content/special-topics/mem-performance.md
```diff
@@ -90,17 +90,27 @@ If you want to make maximal use of your GPU memory for OpenCL, you have three op
...
 - Set [preferences > processing > cpu / gpu / memory > tune OpenCL performance](../preferences-settings/processing.md#cpu--gpu--memory) to "memory size", which will use all of your device's memory, less a 400MB headroom. Please see the [section below](#id-specific-opencl-configuration) for other options related to this setting.
 
+## balanced OpenCL versus CPU tiling
+
+On many machines we find pretty fast multicore cpus with a lot of system ram but a significantly smaller gpu memory with somewhat restricted number crunching power (like typical integrated graphics). There is no problem using OpenCL code in modules as long as that doesn't lead to excessive tiling with large overlapping areas because of large requirements.
+In such cases it might be better to run untiled cpu code instead of heavy tiling on your gpu.
+
+While processing the pipeline we estimate overall workloads for both OpenCL and CPU (this is not static but depends on parameters set in the module thus defining the tile size, number of tiles and tile overlap.
```
gi-man, line 98: This is missing a )

content/special-topics/mem-performance.md
```diff
@@ -90,17 +90,27 @@ If you want to make maximal use of your GPU memory for OpenCL, you have three op
...
+
+In most cases we would prefer the OpenCL code to be executed even if that would mean tiling as the graphics cards is mostly much faster (often 10 times faster if it is a dedicated card).
+If the ratio of estimated workloads (CPU vs GPU) is larger than the **advantage** hint, dt will use the CPU for processing that specific module.
```
gi-man, line 101: I would suggest using darktable instead of dt

elstoc:
I've pushed a copy-edit. Please can you review.

jenshannoschwalm:
Seems good to me.

elstoc:
Thanks! Let's merge

PR #545, opened 2022-12-26
Removed incorrect statement about pushing mouse buttons

colin-adams:


elstoc:
Why is this incorrect? Please can you provide some further explanation or an example.

colin-adams:
The previous paragraph says that you can assign mouse buttons to actions. And indeed I have two of my mouse buttons assigned to apply styles. (not the left or right buttons, but ones that go by the name of tablet button 1 / 2 (or something like that). Despite the names, these are mouse buttons.

colin-adams:
image

elstoc:
Well then let's be more precise here, instead of removing the statement entirely. The left, right, middle and scroll mouse actions are used by the UI and cannot be assigned, but additional buttons can be. We should say that.

It's not a statement that "you can't use buttons" but that you can't use buttons to _initiate_ an action. With the exception of the "tablet buttons" you refer to, you must do something else _before_ clicking a mouse button or else you're just trying to interact directly with the UI.

elstoc:
Mouse buttons are most commonly referred to as left, right, middle, rather than 1,2,3

elstoc:
Thanks!

PR #550, opened 2023-01-13
No xmp read on copy and import

gi-man:
Make it very clear that the existing XMP is not read or used when the user uses Copy & Import.


paperdigits:
Is this true for both light room xmps and dt xmps?

Can we say "Existing XMP files will not be imported."?

gi-man:
> Is this true for both light room xmps and dt xmps?

I could not find lightroom raw + xmp online to test. If one is available, I could test it. I took a darktable sidecar and modified it to filename.xmp format (similar to lightroom). Using Copy & Import, the xmp was ignored. 

I will tweak the wording as requested.

elstoc:
LGTM

Issue #555, opened 2023-02-21
content/special-topics/program-invocation/darktable.md - clarification of info on defaults


gordhub:
The default for --library is described as "$HOME/.config/darktable/library.db". The location for --conf  is described as "$HOME/.config/darktable/darktablerc". These statements are only correct if the the value for --configdir is left at its default "$HOME/.config/darktable/".

I suggest the following to clarify:

1. In both the syntax listing and the description of parameters, move --configdir ahead of --library, so the reader does not need to depend on information not yet presented.
2. In the description of --library, change "The default location of that database file is $HOME/.config/darktable/library.db." to "The default location of that database file is file name library.db in the directory specified or defaulted by --configdir."
3. In the description of --conf, change "darktable supports a rich set of configuration parameters defined by the user in $HOME/.config/darktable/darktablerc." to "darktable supports a rich set of configuration parameters defined by the user in file darktable  in the directory specified or defaulted by --configdir."

gordhub:
Sure...any "house rules" for PRs in addition to the repo's README and the "contributing to dtdocs" page in the manual?

paperdigits:
Rule 1. Do your best.
Rule 2. You'll probably get copy edited no matter how good you do, so don't take offense ;P

PR #556, opened 2023-02-23
 content/special-topics/program-invocation/darktable.md - clarificati…

gordhub:
content/special-topics/program-invocation/darktable.md - clarificaton of info on defaults #555

Changed the descriptions of --library and --conf to be correct in the case where --configdir has been specified.


paperdigits:
Thanks!

PR #558, opened 2023-03-22
Update and expand “setting up opencl” section

sarunasb:
Proposal for updating **setting up opencl** section. Adding instructions for AMD, Intel and Nvidia on Ubuntu 22.10.

### DIFF
--- a/content/special-topics/opencl/setting-up.md
+++ b/content/special-topics/opencl/setting-up.md
@@ -6,55 +6,125 @@ draft: false
 author: "people"
 ---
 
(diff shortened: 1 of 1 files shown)


sarunasb:
> This pull request has not had any activity in the past 60 days and will be closed in 365 days if not updated. Please verify it has no conflicts with the master branch and rebase if needed. Please add a comment if you need help or give permission to other people to finish your work.

Adding some “activity”.

elstoc:
I have discussed this with the other maintainer and we agree that, while this information does seem useful, it is a bit too specific to belong in the user manual and reads more like a "how to" guide. We prefer to keep the manual as a reference and for it to concentrate more on how darktable works so we try to avoid "how to" guides where possible. Also, neither of us are able to verify or maintain any instructions concentrated on a particular Linux distribution. I would suggest that this sort of information might be more appropriate for a site like discuss.pixls.us. Should you choose to post it there (or elsewhere), we would be happy to link to it from within the manual so that it can be discussed/maintained independently.

sarunasb:
OK.

PR #560, opened 2023-03-25
Add indication that color/in and color/out directories must be created by user

gordhub:


elstoc:
Feels a bit like putting "contents may be hot" warnings on coffee cups (do people really not know that if a directory isn't there they must create it before they can put things into it?) but I guess it does no harm. Thanks!

gordhub:
Hilarious!

PR #567, opened 2023-04-17
Document the new options in watermark module

m-p-y:
Document the new options in the watermark module as per PR darktable-org/darktable#13978, which was added to the 4.4 milestone.


elstoc:
Thanks

PR #571, opened 2023-05-07
Mention git in the lua scripts installer docs, add a link to the documentation of the scripts.

kofa73:
- Users often do not know what scripts are available, so I added a link.
- The need to have git installed was only mentioned in the docs of the script manager (https://docs.darktable.org/lua/stable/lua.scripts.manual/scripts/tools/script_manager/#additional-software-required - and even there, the link is wrong -> I've opened a separate PR for that).


elstoc:
Thanks

PR #573, opened 2023-05-21
document additional substitutable variables

ralfbrown:
Prodded by a post on pixls.us, I went through and added available but undocumented variables (except HOURS.AMPM which seems a bit useless since you can't add minutes or seconds in am/pm notation - we need separate 12-hour hours and am/pm marker variables).


elstoc:
Nice one. Thanks @ralfbrown

PR #578, opened 2023-07-02
Update processing.md - remove reference to non-existing setting "auto…

arigit:
…-apply chromatic adaptation defaults"

Update processing.md - remove reference to non-existing setting "auto-apply chromatic adaptation defaults", and add context about the auto-enablement of the "Color Calibration" with "scene-referred" workflow, and how the new handling of White Balance is done in this workflow.

This fixes https://github.com/darktable-org/dtorg/issues/231


elstoc:
This needs more detail now, and perhaps a bit of a rework, since there's a new option for sigmoid and the only option of the four that uses the old chromatic adaptation setting is now (iirc) "legacy" with all others using "modern". Perhaps use the comments in the [release notes](https://github.com/darktable-org/darktable/blob/57f0f678163ebee323c342c336514033c41442c2/RELEASE_NOTES.md?plain=1#L91) as a guide

arigit:
@elstoc thanks for the guidance, I reworked the whole "auto-apply pixel workflow defaults" section, to align it with the release notes, and resubmmitted the PR - please take a look

elstoc:
I think this still needs a note that color calibration is not enabled when selecting "none" but I'll pick that up in the copy-edit. Thanks!

PR #582, opened 2023-08-16
add note that Surface Blur module can be used for color moire

ralfbrown:
As demonstrated by "After Eight Photography" in https://www.youtube.com/watch?v=OeV6DqRWx90


paperdigits:
Thanks!

PR #587, opened 2023-09-13
Make it more clear where we mean a pattern and not a literal

victoryforce:
Here, the `_nn` fragment does not mean a literal doubled letter "n", but is a pattern used to indicate that there should be a number at this place. Therefore, it is more logical to write it in brackets (such as `<nn>`, or even better `<number>`) to convey the information that this is not a literal, but a pattern.


elstoc:
Sounds reasonable to me

PR #588, opened 2023-09-23
Update shortcuts.md with simple hint

dterrahe:
Not sure if that's the correct way to link within the same page...


elstoc:
Thanks!

PR #593, opened 2023-11-23
Add note about conflicting TCA corrections.

paolodepetrillo:
raw chromatic aberrations shifts pixels based on observed TCA in the image content, and then if it's also enabled, lens correction shifts pixels further based on a fixed model, resulting in an over-correction.


paperdigits:
Thanks!

PR #595, opened 2023-11-30
Update darktable-cli.md export options

kmilos:


kmilos:
You can perhaps cross-reference relevant `src/imageio/format/*.c` files w/ `plugins/imageio/format/*` entries in `data/darktablexonfig.xml.in`

PR #601, opened 2023-12-09
Update multiple opencl devices

jenshannoschwalm:
- use more up-to-date so more common cards
- make sure we use the canonical names as that is correct
- update the mentioned pixelpipes as we have a fifth (for second window) since pretty long


elstoc:
LGTM

PR #616, opened 2024-01-31
A few word spelling corrections

victoryforce:


victoryforce:
There is one generally accepted abbreviation for this term (MP). I saw in the texts both an abbreviated and an unabbreviated version of the term. I don't know which one is stylistically better. But if we don't abbreviate it, then I'm not sure which spelling is correct, because all the combinations of the following variants are found in the texts on the web:
- someone writes in the singular, someone in the plural
- someone writes as a capitalized word, someone does not
- someone writes with a hyphen after the number, someone with a space

Therefore, I'd rather delete the last commit, and you fix the uncommon abbreviation as you think is the most correct in English.

elstoc:
For most things, I'd go with an abbreviated term, but MP is (IMO) not the most obvious or commonly used abbreviation and I'd therefore prefer the unabbreviated term. From an English usage point of view a "24 megapixel sensor" is the correct usage. The alternative "24 megapixels sensor" does not make sense and to use the plural you'd need to modify the sentence to something like "a sensor with 24 megapixels". Up to you if you want to correct here or remove the commit, but I'd rather not use "MP"

victoryforce:
> From an English usage point of view a "24 megapixel sensor" is the correct usage.

Great, I'll change the commit accordingly then.

PR #620, opened 2024-02-02
Currently, the default theme is darktable-elegant-grey

victoryforce:
After accepting https://github.com/darktable-org/darktable/pull/11936 the default theme is `darktable-elegant-grey`.

### DIFF
--- a/content/preferences-settings/general.md
+++ b/content/preferences-settings/general.md
@@ -48,7 +48,7 @@ create theme tweaks
 
 ## understanding darktable's themes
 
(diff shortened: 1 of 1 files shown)


elstoc:
On second thoughts this change is not correct. While the default theme _is_ `darktable-elegant-grey`, the bulk of the css remains in `darktable.css`, so this needs some rewording as well.

victoryforce:
I see. Indeed. What would be your suggestion for change? I think we could just remove the phrase about where most of the css code is located.

elstoc:
Since this section is more about how the themes are constructed, I might be tempted to go the other way - i.e. remove the bit about the default theme and change it to say that darktable.css is where most of the css code is located.

victoryforce:
You are right. Done.

elstoc:
LGTM

PR #637, opened 2024-02-16
Reflect the renaming of the shape group's headline

victoryforce:
See https://github.com/darktable-org/darktable/issues/14183.


elstoc:
I'm going to leave this one open for now, as it also needs a change to the screenshot

PR #641, opened 2024-03-04
Fix double space in two md files

lebmich:


lebmich:
Only changes in both .md files should be considered

elstoc:
Then please update your PR so that only these files are included

paperdigits:
You might want to start this PR over, we don't modify PO files or theme files directly. 

PO stuff is handled by weblate and theme files are their own got repo, handled here by submodules.

PR #644, opened 2024-04-09
add cross-ref to rgblevels for 'brightness' (gamma) adjustment in c-b-s

ralfbrown:


elstoc:
I'm not sure whether we want to suggest using rgb levels or curves do we? While they're both "rgb" IIRC the controls/interfaces are still display-referred and I think we'd prefer to direct people to scene-referred modules where possible.

ralfbrown:
True enough, but RGB levels is simpler to use for the cases where the data is within 0...1 (or even a *little* beyond, since it extrapolates the gamma curve).  To properly handle values well above 1, you need to adjust the white and/or gray fulcrum values in colorbalance RGB.  Plus levels gives you a picker to set the gamma from a midtone pixel, which colorbalance doesn't.

elstoc:
Perhaps a rewording then to make it clear that color balance rgb is _recommended_ but that rgb levels provides _similar_ controls to those in the deprecated module.

PR #652, opened 2024-05-23
Use American spelling of "favorite"

victoryforce:


elstoc:
Actually, IIRC the darktable ui did at that point (in v3.4) use the English spelling, which is why I had kept it in the manual. I wonder if it might be worth revisiting that section entirely, it being nearly four years since v3.4.

victoryforce:
> I wonder if it might be worth revisiting that section entirely, it being nearly four years since v3.4.

This section has not been revised yet. Maybe it would make sense to just change the spelling for now?

paperdigits:
That's fair.

elstoc:
Pushed my own commit to do this since it seemed easier than resolving conflicts in this PR

PR #653, opened 2024-06-05
color calibration: make note of sticky color mapping mode and new "ac…

ralfbrown:
…tive" text

Documentation for darktable-org/darktable#16810.


elstoc:
Nice one, thanks @ralfbrown

PR #657, opened 2024-06-22
Update curves.md

alikstimix:
Test case for improvements to text in the DT manual

### DIFF
--- a/content/darkroom/processing-modules/curves.md
+++ b/content/darkroom/processing-modules/curves.md
@@ -44,8 +44,8 @@ scale for graph
 
 : The scale allows you to distort the graph display so that certain graphical properties emerge to help you draw more useful curves. Note that the scaling option only affects the curve display, not the actual parameters stored by the module.
 
-: By default, a “linear” scale is used (defined by a scale factor of 0). This scale uses evenly spaced abscissa and ordinates axes.
+: By default, a “linear” scale is used (defined by a scale factor of 0). This scale uses evenly spaced x and y axes.
 
-: A logarithmic scale will compress the high values and dilate the low values, on both the abscissa and the axis of ordinates, so that the nodes in lowlights get more space on the graph and can be controlled more clearly.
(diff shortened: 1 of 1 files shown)


content/darkroom/processing-modules/curves.md
```diff
@@ -44,8 +44,8 @@ scale for graph
 
 : The scale allows you to distort the graph display so that certain graphical properties emerge to help you draw more useful curves. Note that the scaling option only affects the curve display, not the actual parameters stored by the module.
 
-: By default, a “linear” scale is used (defined by a scale factor of 0). This scale uses evenly spaced abscissa and ordinates axes.
+: By default, a “linear” scale is used (defined by a scale factor of 0). This scale uses evenly spaced horizontal and vertical axes.
 
-: A logarithmic scale will compress the high values and dilate the low values, on both the abscissa and the axis of ordinates, so that the nodes in lowlights get more space on the graph and can be controlled more clearly.
+: The 'scale for graph' slider sets logarithmic scales which will compress the high values and dilate the low values on both the horizontal and vertical axes so that nodes in lowlights get more space on the graph and can be controlled more clearly.
```
elstoc, line 49: This is already under a "scale for graph" heading so there's no need to repeat the name here. I'd perhaps go for something like "Values greater than zero give the graph a logarithmic scale, compressing high values and dilating low values on both..."

alikstimix: Agreed - I did think about merging the 2nd paragraph into 1 but your wording is fine.

alikstimix: Now - how do I go from this discussion page to make the final changes?

elstoc: Don't worry - I'll merge and then update myself. I assume you're using the github web interface to do your PRs. If you intend to do this sort of thing more often I would highly recommend doing it "properly". At a very high level, this requires "forking" the repository, making the changes on a local branch (on your machine), pushing those changes to your fork, and then using the web interface to create the PR. You'll need to do some reading...

PR #658, opened 2024-06-22
add quick reference for actions

ralfbrown:
Start of a "how to" section to point users at modules for accomplishing specific tasks.


elstoc:
I think this is generally a good idea, but will need some thought before I decide to merge. A few minor points in the meantime...

- I won't merge this until all of the modules it references have been documented
- This maybe belongs in the guides/tutorials section
- Will need to consider structure, especially if we think this will grow in the future
- Probably should be linked from the Overview section
- Section headers mostly use lowercase in the user manual, and these sections should too

ralfbrown:
I expect that a lot of the entries will be like hsl.md, just one sentence with a link to the relevant module.  The longer ones like sky replacement might indeed fit better under tutorials, though what I had in mind was a quick reference rather than detailed instructions.

I think we should avoid combining too many things, or people will have trouble finding what they want because it's under some umbrella term in the section's index.

I realized right after pushing that I had automatically capitalized the section headers.  Will fix when revising, once the overall direction is decided.

elstoc:
Just a heads-up on this one... As time goes on I'm less and less convinced we want to do this. Maintenance of this repo is already nobody's top priority and I think this repeats too much information that's available elsewhere.

I think we might need to just concentrate on the absolute minimum required for reference purposes and scale back on big changes like this.

If you're really set on publishing something like this, it might be better as a pixls article or a blog post on dtorg?

ralfbrown:
My gut preference was for a dtorg blog post, and when I looked at the repo I saw that posts are set up the same was as the documentation, which makes the conversion easier.  Do you know of the dtorg blog posts allow nested pages?  (That would make the conversion triival.)  Second choice would be if there's markdown syntax to generate \<details\> tags for collapsible sections, or if any explictly-written ones are preserved.  Last choice would be a list of links to in-page anchors.  I don't think having a completely flat article would be very useful, as it makes quick lookup of specific actions/effects of interest much slower.

elstoc:
The only thing I can suggest is to try it out and see what you can do. The dtorg site also uses hugo under-the-hood, though links probably work differently (we have a custom link conversion in place on dtdocs). It should be fairly easy to set up locally and play, and you could drop into the darktable IRC channel for a chat if you get stuck, as there might be something we could do to make it easier.

ralfbrown:
See https://github.com/darktable-org/dtorg/pull/260

PR #661, opened 2024-07-15
MInor updates to Overiew text

alikstimix:
Latest attempt to use git to update dtdocs

### DIFF
--- a/content/map/overview.md
+++ b/content/map/overview.md
@@ -6,27 +6,27 @@ draft: false
 author: "people"
 ---
 
-The map view allows you to see where your geo-tagged images were taken, and to add location information to non-geo-tagged images.
+The map view shows a world map with all geo-tagged images in the current [collection](../lighttable/digital-asset-management/collections.md) pinned to their geo-tagged location. 
 
-The map view shows a world map with the images in the current [collection](../lighttable/digital-asset-management/collections.md) pinned to their geo-tagged location (if available). This requires that images are tagged with location information. Some newer cameras, including smartphones, are already equipped with GPS receivers. Other cameras may need additional GPS hardware to do this.
+Some newer cameras and smartphones, are already equipped with GPS receivers and can geo-tag each image as the shot is taken. Other cameras may need additional GPS hardware to do this.
 
 Even if your camera doesn't support this feature, there is an alternative method -- darktable can match the Exif date/time in your image(s) to a separate GPX data tracking file created by a GPS tracker, that has recorded your movements. GPS trackers can be purchased as standalone handheld devices or you can install a GPS tracker app on your smartphone. Location tagging with GPS tracking data can be done using the [geotagging](../module-reference/utility-modules/shared/geotagging.md) module, in the lighttable and map views.
 
+In addition, you can assign geo coordinates to an image, by activating the [filmstrip](../module-reference/utility-modules/shared/filmstrip.md) on the lower panel (press Ctrl+F)and dragging the image icon from the film-strip to a position on the map.  Darktable will record the new location (latitude and longitude ) as part of the image metadata and this data will be included in exported images.
+
 # center map view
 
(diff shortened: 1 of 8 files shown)


content/map/overview.md
```diff
@@ -6,27 +6,27 @@ draft: false
...
+Some newer cameras and smartphones, are already equipped with GPS receivers and can geo-tag each image as the shot is taken. Other cameras may need additional GPS hardware to do this.
 
 Even if your camera doesn't support this feature, there is an alternative method -- darktable can match the Exif date/time in your image(s) to a separate GPX data tracking file created by a GPS tracker, that has recorded your movements. GPS trackers can be purchased as standalone handheld devices or you can install a GPS tracker app on your smartphone. Location tagging with GPS tracking data can be done using the [geotagging](../module-reference/utility-modules/shared/geotagging.md) module, in the lighttable and map views.
 
-# center map view
-
-In the center of the map view you will see a world map.
+In addition, you can assign geo coordinates to an image, by activating the [filmstrip](../module-reference/utility-modules/shared/filmstrip.md) on the lower panel (press Ctrl+F)and dragging the image icon from the film-strip to a position on the map.  Darktable will record the new location (latitude and longitude ) as part of the image metadata and this data will be included in exported images.
```
paperdigits, line 15: Is it "filmstrip" or "film-strip" ?

content/map/overview.md
```diff
@@ -6,27 +6,27 @@ draft: false
...
 
-Map data is taken from open map sources on the internet. As such, new map data is only available if you are connected to the internet -- darktable keeps a disk cache of previously loaded map data.
+# center map view
 
-You can navigate within the map using your mouse. Left-click and drag to move the map. Use the scroll-wheel to zoom in and out.
+In the center panel you will see a world map sourced from an Internet provider.  Darktable keeps a disk cache of previously loaded map data so new/updated map data is only available when you are online.  
+You can choose your provider from:
+OpenStreetMap, Maps-For-Free, OpenCycleMap, Public Transport, Google Maps, Google Satellite, Google Hybrid, Virtual Earth, Virtual Earth Satellite, Virtual Earth Hybrid.
```
paperdigits, line 21: I don't think we should list these, they're obvious and subject to change with the underlying library.

content/map/overview.md
elstoc: filmstrip

content/map/overview.md
```diff
@@ -6,27 +6,27 @@ draft: false
...
-# center map view
-
-In the center of the map view you will see a world map.
+In addition, you can assign geo coordinates to an image, by activating the [filmstrip](../module-reference/utility-modules/shared/filmstrip.md) on the lower panel (press Ctrl+F)and dragging the image icon from the film-strip to a position on the map.  Darktable will record the new location (latitude and longitude ) as part of the image metadata and this data will be included in exported images.
 
-Map data is taken from open map sources on the internet. As such, new map data is only available if you are connected to the internet -- darktable keeps a disk cache of previously loaded map data.
+# center map view
 
-You can navigate within the map using your mouse. Left-click and drag to move the map. Use the scroll-wheel to zoom in and out.
+In the center panel you will see a world map sourced from an Internet provider.  Darktable keeps a disk cache of previously loaded map data so new/updated map data is only available when you are online.
```
victoryforce, line 19: Definitely, a world map is NOT sourced from an Internet provider. An ISP only provides access to the Internet where another organization provides a mapping service.

alikstimix: Victor, Internet provider does not equal ISP, but I take your point that users could be confused so would you agree with this text:

"In the center panel you will see a world map sourced from an Internet based mapping service"

victoryforce: > Victor, Internet provider does not equal ISP

Really??? Can you provide some proof of this statement?

PR #662, opened 2024-07-15
Update content/print/overview.md

alikstimix:
Simple update to test my understanding of the process.

### DIFF
--- a/content/print/overview.md
+++ b/content/print/overview.md
@@ -6,7 +6,7 @@ draft: false
 author: "people"
 ---
 
-This view allows you to print your images. Because printing is not easy, there are many technical aspects to be taken into account.
+This view allows you to print your images. 
 
 After selecting an image in the [lighttable view](../lighttable/_index.md) you can enter the [print settings](../module-reference/utility-modules/print/print-settings.md) module to adjust printer settings and initiate printing.
 


content/print/overview.md
```diff
@@ -6,10 +6,10 @@ draft: false
 author: "people"
 ---
 
-This view allows you to print your images. Because printing is not easy, there are many technical aspects to be taken into account.
+This view allows you to print your images. 
 
-After selecting an image in the [lighttable view](../lighttable/_index.md) you can enter the [print settings](../module-reference/utility-modules/print/print-settings.md) module to adjust printer settings and initiate printing.
+Darktable only scans for printers at startup, so your target printer must be active when darktable starts in order to provide the print module with the target printer's ICC profile.
```
victoryforce, line 11: Did you test it yourself?

alikstimix: I did!  Several times!!  

I have a colour printer that is normally switched off as I rarely print anything.  When I first looked at the Print view I couldnt load any image into centre panel, so nothing could work.  After some head scratching and checking of Issues in gihub, I managed to understand why.

If I turn my printer on, and let it settle down I can then start dt and see my printer in the dropdown list.  Of course I still have to know to select an image in lighttable after which I get the image in Print view, plus drag and drop from the filmstrip.  And I can print out a colour image just fine.

alikstimix:
> That's strange, certainly not something I see on my side. I'm away from home at the moment and see my printer in the list and I can select it.

Its on the Issue list as #15575

alikstimix:
Actually I just tested against my wife's printer, and now it all works fine.

I'll investigate more tomorow.

paperdigits:
IMO there isn't really a need to be that specific.

alikstimix:
OK I've pinned it down repeatably.

Start with a new instance of dt, and nothing works until you 
a)select and image in lightable
b) double click on a printer in the print right panel

After that it seems everything is cached and will work as expected.

Mica - I agree the manual doesnt need to be this specific, unless this is a bug/expected behaviour that won't be changing anytime soon.

elstoc:
We don't document bugs in the manual, only expected behaviour. If the printer is expected to be in the list at all times, then anything else is a bug. BTW my laser printer is off at the moment and I can see it and select it in the print view.

paperdigits:
I'd expect much of this is about cups and how it is configured by the distro or package, and in that case being less specific is preferable, if things change from distro to distro or version to version we don't have to update it. Updates to the English text is magnified n-times by the number of translations.

alikstimix:
That's a good point about translation magnifying the workload after a change to the English text.

There are lots of sections that could be improved a little, but it has to be a big improvement to justify 7 translations.

content/print/overview.md
```diff
@@ -6,10 +6,8 @@ draft: false
...
 
-This view allows you to print your images. Because printing is not easy, there are many technical aspects to be taken into account.
+This view allows you to select and print your images.
 
-After selecting an image in the [lighttable view](../lighttable/_index.md) you can enter the [print settings](../module-reference/utility-modules/print/print-settings.md) module to adjust printer settings and initiate printing.
+The first step is to select an image in the [lighttable view](../lighttable/_index.md), then switch to the print view. The selected image will be displayed in the centre panel and the [print settings](../module-reference/utility-modules/print/print-settings.md) module will now allow you to adjust printer settings and initiate printing.
 
-This module supports the printer's ICC profile, which is somewhat mandatory if you want to obtain a high quality print close to the image displayed on the screen.
-
-It is important to note that ICC profiles provided by the paper and/or printer manufacturers cannot be used on GNU/Linux as they are printer-driver dependent. The darktable print module uses CUPS and there are no ready-to-use ICC profiles available for this driver.
```
elstoc, line 15: Is there a reason for removing this "important" paragraph?

alikstimix: Well the text is pretty opaque, and its hard to see what relevance it  has to the user.  It seems to be describing an aspect of the workflow of the module, and is not something the user can affect.

Maybe its of use to whoever is maintaning the module, but even then it needs a rewrite.

content/print/overview.md
victoryforce: Examining the source code shows that darktable does not scan printers at startup. Therefore, your statement is incorrect.

content/print/overview.md
elstoc: It looks to be useful information for users and I'm pretty sure I understand what it's saying (you cannot use manufacturer-produced ICC profiles because they're designed for the manufacturer-produced printer drivers -- i.e. for Window$ and/or Mac) . If _you_ don't understand it, that's not necessarily a reason to remove it.

alikstimix: Chris,  You and I may understand it, but we are not the target audience for the user manual.

Users come to the print overview because they want to understand dt printing better, or have a particular problem.  Therefore the text should be accessible to an intelligent non-specialist.  

If the user doesn't already know what an ICC profile is, then the text in question is unlikely to be relevant to their problems, and the warning is superfluous.

If they do know a bit about ICC profiles then, in my opinion, the text is imprecise and in some ways self contradicting.

"this module **supports** the printer's ICC profile . . ." probably means "this module **uses*  . . . "  and the rest of the sentence just states the obvious.

The sentence about not using the printer ICC profile on Linux is hard to understand.  I only run Linux , so where does the profile fit in to my print workflow?  Also, dt does not provide any means to directly access the printer profile, so what is the text warning against?  If the user really wants to try to reuse the printer ICC profile, they must already have some knowledge and expertise outside of dt in order to get hold of it.  In which case they wouldnt be asking the dt manual for help.

So anyway, my feeling is that the text would be much clearer and potentially more useful if it included a description of the dt workflow for printing - explaining why multiple ICC profiles are needed and how dt interacts with printer drivers.  In this case, I don't think the text belongs in a section called "overview" but potentially in a new section called Print Workflow, that can include more technical information.

Overall I am just providing feedback and suggestions, you are the gatekeeper - your call!

PR #663, opened 2024-07-16
Update views.md

alikstimix:
Update text to reflect the existence of the "Other" dropdown

### DIFF
--- a/content/overview/user-interface/views.md
+++ b/content/overview/user-interface/views.md
@@ -6,13 +6,21 @@ weight: 10
 author: "people"
 ---
 
-The functionality in darktable is separated into six different views:
+The functionality in darktable is delivered through two main views and four subsidiary views:
+
+**Main Views**
+<br>Switch between main views by clicking the view name at the top right of the window:
 
 [lighttable](../../lighttable/_index.md)
 : Manage images and collections.
 
 [darkroom](../../darkroom/_index.md)
 : Develop a single image.
+
+The currently active view is highlighted.
+
+**Other Views**
+<br>Switch to subsidiary views by choosing from the "Other" dropdown at the top right of the window:
 
 [map](../../map/_index.md)
 : Show geo-tagged images on a map and manually geo-tag new images.
@@ -26,7 +34,7 @@ The functionality in darktable is separated into six different views:
 [tethering](../../tethering/_index.md)
 : Remotely capture and save images taken with a connected camera.
 
-You can switch between views by clicking the view name at the top of the right-hand panel (the currently active view is highlighted) or by using one of the following keyboard shortcuts:
+You can also switch between views by using one of the following keyboard shortcuts:
 
  - `L` switches to lighttable
  - `D` switches to darkroom


content/overview/user-interface/views.md
```diff
@@ -6,14 +6,22 @@ weight: 10
 author: "people"
 ---
 
-The functionality in darktable is separated into six different views:
+The functionality in darktable is delivered through two main views and four subsidiary views:
+
+**Main Views**
```
elstoc, line 11: I think this should use markdown "header" syntax (with the `#` symbol) rather than just bold text. Additionally dtdocs has headings in lower case to match the style of the application UI.

content/overview/user-interface/views.md
```diff
@@ -6,14 +6,22 @@ weight: 10
...
+<br>Switch between main views by clicking the view name at the top right of the window:
```
elstoc, line 12: Please avoid html (remove the `<br>`). If you must have a new paragraph you can use a line feed in markdown. Similarly below.

content/overview/user-interface/views.md
```diff
@@ -6,14 +6,22 @@ weight: 10
...
 
 [lighttable](../../lighttable/_index.md)
 : Manage images and collections.
 
 [darkroom](../../darkroom/_index.md)
 : Develop a single image.
 
+The currently active view is highlighted.
+
+**Other Views**
```
elstoc, line 22: Let's keep the terminology consistent. Are they subsidiary views or "other" views? Given that the application uses the term "other" I'd go with that, but I agree it makes the text harder to write. On reflection I'm not sure it's worth the effort to call them "main" and "other/subsidiary". Perhaps it's better to just add a line before the map/tethering sections saying something like "the following views are available by clicking on the 'other' label"

elstoc (commented):
Reading this I'm not sure it's worth drawing a distinction between "main" and "subsidiary" views

content/overview/user-interface/views.md
alikstimix: Agreed

content/overview/user-interface/views.md
alikstimix: Agree with using "other" instead of subsidiary.

The main reason for splitting out the "Other" views is that the UI already casts them as different/lesser by relegating them to the dropdown.  So the manual should emphasise the ease of switching between lightable and darkroom which is where users will spend 90+% of their time.

Your call.

alikstimix: On a practical note. I can make the corrections to the text in my fork alixtimix/dtdocs/patch-4 

Once I commit those changes, how do I send the updated material back to you?  A new push request? or what?

elstoc: If you update your branch that automatically updates the PR

alikstimix: OK all done

paperdigits:
Please use markdown headers, not bolded text.

content/overview/user-interface/views.md
paperdigits: also note it isn't always called "other" -- when you change to the map view, then it is called "map", etc etc. So what it is called depends on what view you're currently in.

PR #664, opened 2024-07-16
Update rendering-intent.md

alikstimix:
Added short explanation of what a rendering intent is.

### DIFF
--- a/content/special-topics/color-management/rendering-intent.md
+++ b/content/special-topics/color-management/rendering-intent.md
@@ -5,6 +5,7 @@ weight: 40
 draft: false
 author: "people"
 ---
+A rendering intent is a method of converting colours from one ICC profile to another.They handle issues when converting from one profile to another, as there may be colours that are in gamut of the source profile but out of gamut of the destination.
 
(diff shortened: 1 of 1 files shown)


content/special-topics/color-management/rendering-intent.md
```diff
@@ -5,6 +5,7 @@ weight: 40
 draft: false
 author: "people"
 ---
+A rendering intent is a method of converting colours from one ICC profile to another.They handle issues when converting from one profile to another, as there may be colours that are in gamut of the source profile but out of gamut of the destination.
```
paperdigits, line 8: Needs a space after the first period.

paperdigits:
I am not sure we should be explaining rendering intent to people... it isn't really the place of the manual to explain general concepts to the user.

elstoc:
The page already defines what the rendering intent is, and it's nothing to do with ICC profiles: "define how to handle out-of-gamut colors when converting between color spaces"

alikstimix:
When I first started using dt, I had no idea about rendering intents, but I could see that a choice had to be made even when exporting an image to jpg.  My feeling is that users in that situation should be able to go to the "rendering intent" page of the dt manual and see a short definition of rendering intent.  

Chris - the existing text really does not give a definition, and looking at half  a dozen web pages that do give a definition, they all link rendering closely with ICC profiles. I don't claim to be a colour management expert, but the text I proposed seems to be in line with most explanations I can find.

elstoc:
I would rather wording like this be confirmed by someone who _does_ know what it means because It doesn't look right to me. I'm sure the term rendering intent _can_ be used in the context of ICC profiles, but I'm also pretty sure that's not the intended context here.

PR #668, opened 2024-07-20
Initial Color Eq docs based on documentation by Boris.

paperdigits:
Color Eq docs based on a document sent to me by Boris. 

darktable issues that might need labels adjusted if this is merged: 
https://github.com/darktable-org/darktable/pull/16633
https://github.com/darktable-org/darktable/pull/16543
https://github.com/darktable-org/darktable/pull/16446
https://github.com/darktable-org/darktable/pull/16400
https://github.com/darktable-org/darktable/pull/16301
https://github.com/darktable-org/darktable/pull/16279
https://github.com/darktable-org/darktable/pull/16236
https://github.com/darktable-org/darktable/pull/16089


s7habo:
Hi Mica,  here I am.  How do we proceed now? I'm almost finished with Manual as a document. Tomorrow I can offer it.

paperdigits:
@s7habo if you click on the "Files Change" tab on this page, you'll see the page I added. You can either add comments, click and edit it to propose changes, or make comments that I can accept as changes. You can see the flow for that here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request

Feel free to add whatever you'd like!

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+Click+drag the control points on the curve to change the saturation of all pixels where the mask includes the given color.
+
+## brightness tab
+
+Click+drag the control points on the curve to change the lightness of all pixels where the mask includes the given color.
+
+## node placement
+
+node placement
+: Move the position of the nodes horizontally.
```
s7habo, line 46:
```suggestion
: Move the position of the nodes horizontally.

-   By moving the node with the left mouse button pressed.
-   Clicking on the node with the right mouse button opens a multi-functional pop-up window where the numerical values can be entered or manipulated by using the arrow keys on the keyboard:

    
![ce_pop_up](https://github.com/user-attachments/assets/37c763d7-8269-4a15-9970-f7fae2b274b2)

- When hovering the mouse cursor over the node  (the node turns white) and scrolling the mouse wheel while holding down the ctrl+shift keys. Finer steps can be achieved by holding down only the ctrl key.

```

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+
```
s7habo, line 47: ## range selection

To better estimate which nodes should be used for manipulation, you can use the color picker to select the exact point or color range in the photo. 

Activate the color picker and click on one spot in the image. A white vertical line indicates where the pixel is located in the hue range:

ce_color_picker

If you activate the color picker by holding down the ctrl key, an area can be selected:

ce_color_picker_2

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+: Use the color picker to show a color from the photo on the equalizer interface.
+
+## options
+
+white level
+: Set the upper bound for the brightness correction via the slider or picker.
```
s7habo, line 53: The default is fine for most images.

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+
+hue curve
+: Control how the curve is interpolated between control points. Move the slider to the right to make the transitions between the control points more gradual. Move the slider to the left for a sharper transition, but beware that this can result in harsher transitions.
```
s7habo, line 56: Note: This only applies to the _hue curve_ and _not_ to the _saturation and brightness_ curves!

With the default (value of 1), the curve looks like this:

ce_hue_curve_1

This is what it looks like with a minimum (value of 0.05):

ce_hue_curve_2

This allows the specific hue to be manipulated more precisely without affecting neighboring areas too much.

elstoc:
FYI we want to try and avoid too many images if possible, since it's a pain to re-create them every time the UI or functionality changes.

s7habo:
@paperdigits  is it ok for you if I comment like this?

I would hate to make sugestion about changes in the code itself because I have no idea how to use this language.

> FYI we want to try and avoid too many images if possible, since it's a pain to re-create them every time the UI or functionality changes.

@elstoc I understand that this means more work, but it also means a lot of work to explain things to users afterwards, when there are hardly any visual examples for functions that are sometimes explained in a very technical way.

I will be happy to re-create them  afterwards if the changes require it.  When the time has come, you are welcome to contact me.

paperdigits:
@s7habo yeah this is fine. If you can "suggest" the changes, then I can just click "accept" instead of having to copy and paste. 

Thanks for the work, the range selection stuff is completely unrecoverable from just the UI itself :D

elstoc:
@s7habo I don't mean _no_ images but let's try to limit the number if we can and make them simple enough that they won't need too much maintenance later on. If we are using images to illustrate concepts, it's better to include only as much of the UI as you need to explain things. For example if you're trying to illustrate the shape of the curve when a parameter is applied, then if you only show the curve itself and not the rest of the module UI, it will probably never need to be updated again.

content/module-reference/processing-modules/color-equalizer.md
elstoc: I think it's good to illustrate point and range selection with color pickers, but this might be better in a more generic section, since it will apply to other modules as well.

s7habo:
> .If we are using images to illustrate concepts, it's better to include only as much of the UI as you need to explain things. For example if you're trying to illustrate the shape of the curve when a parameter is applied, then if you only show the curve itself and not the rest of the module UI, it will probably never need to be updated again.

Ok. I will try to pay attention to it. I will continue now and then we can talk about the details.

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+
```
s7habo, line 57:
```suggestion
## guided filter
Guided filter can help to prevent or at least reduce the harsh overlap between manipulated and non-manipulated areas. It is particularly useful for noisy photos or areas with strong local color variations that could lead to strong local color contrasts or steep color gradations.

Here is a piece of meadow with darkened green grass without…

![ce_guided_off](https://github.com/user-attachments/assets/6b598b31-f285-4202-ba24-257f886fd804)
...and with the guided filter switched on. With the guided filter the texture of the grass looks much softer and calmer:

![ce_guided_on](https://github.com/user-attachments/assets/cd0960ee-aa76-486e-bcf0-67827cb9ba83)


```

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+use guided filter
+: Enable or disable the use of the guided filter to separate each color node.
+
+hue analysis radius
+: Set the radius for the chroma analysis guided filter. Increase the value if there is a large amount of hue variance or chroma noise.
```
s7habo, line 62:
```suggestion

Guided filter that is used for the chroma analysis which pixels should be considered for manipulation. 

The point here is to take into account not only the pixels themselves, but also the _area around them defined by the radius of guided filter_, which ensures that, for example, in the case of very noisy photos or photos with Moiré stripes, the areas between pixels are also taken into account which, in those cases, ensures a more “homogeneous” selection.

Here is an example with a noisy photo of the clothing with Moiré stripes. Hue analysis radius is in its default position (1.5. px):

![ce_moire_1](https://github.com/user-attachments/assets/4538532c-a646-4543-8359-0e4dc51335ce)

With hue analysis radius of 10 px. The Moiré effect is now much less pronounced:
![ce_moire_3](https://github.com/user-attachments/assets/9c7a2c64-52f1-4444-a96c-228ce1482cf1)

With additional denoising it disappears completely:
![ce_moire_2](https://github.com/user-attachments/assets/7bf83652-d970-43f9-addc-05b51fe3074c)

Note: High values can lead to haloing at the edges!

```

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+
+saturation threshold
```
s7habo, line 64:
```suggestion
saturation threshold

The saturation threshold as additional control for the pixel selection. Especially by portraits or photos where we want to enhance the texture, it can be useful to set the saturation threshold when selecting and manipulating pixels.

_Visualization indicator_  
![grafik](https://github.com/user-attachments/assets/be6b7a37-fa9f-4f93-9af4-01c5c05f71a6)
shows which pixels have been selected. Red color indicates the selection, and pixels in the blue area are not taken into account.

Here is an example of the color selection of green leaves based on saturation threshold.

With saturation threshold practically disabled (at 0.0%), visualization indicator shows that the leaves are almost completely selected:

![ce_saturation_treshold_1](https://github.com/user-attachments/assets/92e77378-d9e7-4a8a-9590-75ebba05a9cf)

With a 20% saturation threshold, _only part of the green with higher saturation_ in the leaves _is selected_ (the red area):

![ce_saturation_treshold_2](https://github.com/user-attachments/assets/ca620d6a-2cfc-4914-9cb9-cc1c4acc1e7b)

By manipulating _only that high saturation area_, we have now enhanced the texture of the leaves:

![ce_saturation_treshold_4](https://github.com/user-attachments/assets/3a2d7560-a108-42fd-a023-e4c813b635f2)

```

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+: Set the upper bound for the guided filter's saturation threshold. Decrase the value to allow changes to areas with low chromaticity. Increase the value to restrict changes to areas of high chromaticity.
+
+contrast
```
s7habo, line 67:
```suggestion
contrast

Contrast slider regulates the steepness of the curve defined by saturation threshold:

contrast value -1:

![ce_contrast_1](https://github.com/user-attachments/assets/b4883918-3132-4536-8dac-002690610b14)

contrast value 1. The transition is much sharper. Thin green curve in the visualization indicator represents steepness of transition:

![ce_contrast_2](https://github.com/user-attachments/assets/afb72f21-e7bc-4921-8de0-d6b916b14ba0)

```

s7habo:
The point “effect radius” is still open. I'll deal with that tomorrow.

content/module-reference/processing-modules/color-equalizer.md
s7habo: > I think it's good to illustrate point and range selection with color pickers, but this might be better in a more generic section, since it will apply to other modules as well.

I agree. This is an important point and makes sense in terms of general use. Ideally, all modules that offer area selection should do so in the same way.  Parametric masks and color zones have an extra color picker for this.

Global color picker and color equalizer do it via keyboard shortcut.

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,71 @@
...
+: Set the color contrast for the guided filter. Incrase th value to favor sharp transitions between colors and increase color contrast. Decrease the value for smoother color transition.
+
+effect radius
```
s7habo, line 70:
```suggestion
effect radius

Radius of the guided filter which is used for pixel manipulation to smooth the result while preserving the edges. 
This can be useful for both preserving detail or smoothing the result, depending on the subject and the manipulation  that have been performed.

In this case, we want to lighten the pumpkin. Without the guided filter or with a very low radius, we are loosing the details:

![ce_effect_radius_4](https://github.com/user-attachments/assets/e16f7d9d-e960-40a2-bb49-f884070d8b52)

If we then increase the radius in order to smooth the result, the details are retained:

![ce_effect_radius_5](https://github.com/user-attachments/assets/908744b2-7bb2-4a2f-abe0-9bad8cf14740)

To prevent the “bleeding” at the edge of the pumpkin caused by large radius of the guided filter, we can also increase the saturation threshold to limit the manipulation to the pumpkin, which stands out from the surroundings through saturation:

![ce_effect_radius_6](https://github.com/user-attachments/assets/4806a130-7690-4204-997e-92b8232727c5)

We can also use the visualization indicator to better see which areas have been changed. The pixel areas whose value has been increased are displayed in red and the areas with a reduced value are displayed in blue:

![ce_effect_radius_7](https://github.com/user-attachments/assets/3fe755d4-1fee-4876-a743-bf0db4c5d319)


```

s7habo:
I am now finished. I also have the whole thing as an odt document which I can offer together with screenshots.

content/module-reference/processing-modules/color-equalizer.md
paperdigits: I don't think the screenshots are necessary here.

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,159 @@
...
+![ce_effect_radius_5](https://github.com/user-attachments/assets/908744b2-7bb2-4a2f-abe0-9bad8cf14740)
+
+To prevent the “bleeding” at the edge of the pumpkin caused by large radius of the guided filter, we can also increase the saturation threshold to limit the manipulation to the pumpkin, which stands out from the surroundings through saturation:
+
+![ce_effect_radius_6](https://github.com/user-attachments/assets/4806a130-7690-4204-997e-92b8232727c5)
+
+We can also use the visualization indicator to better see which areas have been changed. The pixel areas whose value has been increased are displayed in red and the areas with a reduced value are displayed in blue:
+
+![ce_effect_radius_7](https://github.com/user-attachments/assets/3fe755d4-1fee-4876-a743-bf0db4c5d319)
+
```
ralfbrown, line 157: As with the saturation threshold, the visualization can only be seen in areas where the blend mask selects the module's output.

jenshannoschwalm:
A BIG thank you to all of you.

BTW I will check how/if the mask visualizing restriction mentioned by @ralfbrown can be avoided.

jenshannoschwalm:
Yes but we have a way to pass thru the RGB data.

content/module-reference/processing-modules/color-equalizer.md
paperdigits: I decided to go minimal and forego the images here too, if you feel strongly we need images, please feel free to add them.

paperdigits:
@elstoc I hope this is good enough for a first draft.

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,156 @@
...
+include_toc: true
 ---
 
+Selectively adjust the hue, saturation, and brightness of pixels based on the color.
+
+Color Equalizer is essentially an implementation of the Color Zones module, but works in the scene referred part of darktable's pixelpipe.
+
+This module works in darktable's UCS 22 color space.
+
+The entire hue range is represented as a curve with eight overlapping sections divided by eight fixed nodes. Choose whether you wish to adjust (select) pixels based on their hue, saturation, and brightness. You can then use the equalizer interface nodes, on their respective tabs, to adjust the hue, saturation, and brightness of ranges of pixels selected via this method.
```
elstoc, line 19: You can't choose to adjust/select by saturation or brightness. All of the selections are by hue. It should perhaps say something like "Pixels are selected based on their hue, and the module can be used to adjust the hue, saturation, or brightness of those selected pixels"

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,156 @@
...
+
+The distances between nodes cannot be changed but all nodes together can be moved by +/- 23 degrees along the color wheel using the node placement slider, allowing you to fine-tune the selection of a specific color.
```
elstoc, line 21: No need to specify `+/- 23 degrees` here

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,156 @@
...
+
+**Note**: Because all processing in color  equalizer is based on the hue data of each pixel, it can result in increased chroma noise or even increased luma noise if a lot of brightness changes are applied.
+
+---
+
+# module controls
+
+## hue tab
+
+Click+drag the control points on the curve to change the color of all pixels where the mask includes the given color.
```
elstoc, line 33: This is the first time we've mentioned the word "mask"

content/module-reference/processing-modules/color-equalizer.md
```diff
@@ -1,7 +1,156 @@
...
+view: darkroom
+masking: true
+include_toc: true
 ---
 
+Selectively adjust the hue, saturation, and brightness of pixels based on the color.
+
+Color Equalizer is essentially an implementation of the Color Zones module, but works in the scene referred part of darktable's pixelpipe.
+
+This module works in darktable's UCS 22 color space.
```
elstoc, line 17: I don't think most users are going to care about this

elstoc:
BTW I've created a separate issue around central documentation of color pickers (https://github.com/darktable-org/dtdocs/issues/673)

elstoc:
As the discussion is getting a bit hard to follow, and I wanted to do a bit of reorganisation, I have created another draft in a new PR #674

PR #674, opened 2024-08-17
Color Equalizer documentation - alternative proposal

elstoc:
Based on the initial draft in #668, this adapts the content to better fit the existing documentation style.

This draft omits the images and some of the linked examples/illustrations, mostly because I'd rather make sure the text makes as much sense as possible without the images before we add them back in.

IMO this content could be re-added later within its own "examples" section at the end of the page, similar to the way the [refinement controls](https://docs.darktable.org/usermanual/4.6/en/darkroom/masking-and-blending/masks/refinement-controls/) section is structured.


elstoc:
Ok, no comments is good comments. Merging...

PR #676, opened 2024-08-28
Remove incorrect warning in negadoctor manual

sayoder:
shadows color cast does indeed have an effect when the scanner exposure bias is non-zero. Perhaps this was different in older DT versions.


elstoc:
This is not an area of functionality that I use, so I'll take your word for it, since the sentence in question probably doesn't change the meaning of the documentation

PR #682, opened 2024-10-03
Bring rating sample workflow explanation up to date

HobbyProgrammer:
In the _import & review_ section, it says that by default, Darktable rates imported images with a one-star rating, and that pressing `0` gives images a zero-star rating. I just started using Darktable. After importing images, I found that they are marked as `Not rated`, and that there is no zero star rating. I guess this part of the manual is outdated. Here is my attempt on bringing it up to date.


elstoc:
I'd like to prove that change of behaviour myself and understand whether it was changed on purpose or was a bug/regression introduced unintentionally. Do you know when that behaviour changed? It would be good to find the darktable PR that changed it

HobbyProgrammer:
Okay, I dug deeper into what is happening for me, now. It seems that my camera puts `Rating 0` into the EXIF data of all images by default. The default settings for import are `ignore EXIF rating = false` and `initial rating = 1`. The description on `ignore EXIF rating` says _if not set and EXIF rating is found, it overrides 'inital rating'_. This explains why I was observing a different behavior than what is explained in the manual. With this, the error is on my end, and we can close the PR, or if you think it is worth it, mention this parameter in the import tab to avoid confusion for future readers. 
Personally, I was also slightly confused with the `zero-star` rating term, since the UI is showing `Not rated`, but I think that is a very minor detail.

paperdigits:
OP says close, we close.

PR #686, opened 2024-11-03
Removed docs of opencl_synch_cache

kofa73:
`opencl_synch_cache` was removed in 4.2, see https://github.com/darktable-org/darktable/pull/12504 It's interesting that the docs were updated, but instead of removal, the following was added:
> Since darktable 4.4 the efficiacy of the pixelpipe cache has improved a lot so setting it to "active module" (the default) is good in most cases.


elstoc:
Thanks!

PR #687, opened 2024-11-04
Fixed description of load sidecar file

kofa73:
Part of the original text was probably left-over from copied from the copy/paste operations, leading to confusion as witnessed by several users on discuss.pixls.us thinking it was possible to load a sidecar file in append mode.


elstoc:
LGTM

PR #688, opened 2024-11-25
spelling and grammar fixes for the Overview area in German

mayonnaisehypernova:
While reading through the darktable docs in German, I noticed some spelling mistakes and partially weird grammar that I wanted to fix. 

This is only for the "Overview" area. I might create more commits / PRs as soon as I have time to continue reading.


elstoc:
I'm not sure what to do with this one. We generally only allow translation changes via weblate, but weblate is currently down. I'll leave this open for now and update once we've discussed the way forward.

paperdigits:
We can take modifications to the po files. When weblate comes back, it'll just pull all the changes in. Weblate is a fancy po editor.

PR #697, opened 2024-12-17
update quick-access-panel screenshot

ralfbrown:
Showcase more of the recently-added widgets, the presets menus, and the reset buttons.


elstoc:
Is there additional functionality in that PR (you mention presets menus and reset buttons above) or is it just adding more default widgets? If there's additional functionality we should probably document that in this PR as well

ralfbrown:
Preset menus and reset buttons may be worth mentioning, though we can probably just copy the descriptions from the module overview page since they function the same.

elstoc:
This has been done under a separate PR now.

PR #712, opened 2024-12-30
dtdoc edit for "darkroom: Add synchronization of last edited iop to selection."

tnbde:
[darktable PR#17589](https://github.com/darktable-org/darktable/pull/17589) 
- The first commit adds in the filmstrip.md an additional item in the shortcut list describing the new shortcut Ctrl+X implemented by the darktable PR metioned above. In addition, I decided to clarify the first sentence, since from my understanding the filmstrip is only available in the darkroom view.
- The second commit is an edit of the last paragraph mentioning the filmstrip module in the darkroom-view-layout.md, so that it is formatted more consistently like all other modules mentioned in this document.


content/darkroom/darkroom-view-layout.md
```diff
@@ -91,4 +91,5 @@ From left to right:
 ![guides-overlays-icon](./darkroom-view-layout/guides-overlays-icon.png#icon) [guides & overlays](../module-reference/utility-modules/darkroom/guides-overlays.md)
 : Left-click to switch global guide overlays on/off and right-click to change the guide settings, including the color of all on-image drawing (masks, crop guides etc.)
 
-You can also enable the [filmstrip](../module-reference/utility-modules/shared/filmstrip.md) module at the bottom of the screen to allow you select and interact with the currently selected collection in the [lighttable](../lighttable/_index.md) view.
+[filmstrip](../module-reference/utility-modules/shared/filmstrip.md)
+: You can also enable the filmstrip at the bottom of the screen allowing you to select and interact with the currently selected collection similar to the [lighttable](../lighttable/_index.md) view.
```
elstoc, line 95: This isn't a formatting error, it's on purpose. The definition list describes the buttons, but filmstrip is not a button. I don't think we should change this.

content/module-reference/utility-modules/shared/filmstrip.md
```diff
@@ -6,7 +6,7 @@ tags:
 view: darkroom, tethering, map, print
 ---
 
-The filmstrip can be used to quickly switch between images. The images shown are the same as those displayed in the lighttable view and are defined by the currently-selected collection.
+The filmstrip can be used to quickly switch between images in the darkroom. The images shown are the same as those displayed in the lighttable view and are defined by the currently-selected collection.
```
elstoc, line 9: This is documented here (under `utility-modules/shared`) because the filmstrip is used in multiple views, not just the darkroom.

elstoc (changes requested):

content/module-reference/utility-modules/shared/filmstrip.md
tnbde: Thanks for the clarification, I only knew the filmstrip module from the darktable view, I will correct it.

content/darkroom/darkroom-view-layout.md
tnbde: One advantage is here, that the filmstrip module is formatted in the same way in this chapter/section, like all other modules as in the left and right panel subsections.

But if it was intended this way, i will correct it. Thanks for the feedback!

elstoc: I do think there's a good argument for filmstrip being a toggle button on the button bar, but that's another discussion

tnbde: I agree. This should be reverted now.

elstoc:
Thanks

PR #714, opened 2025-01-01
Documented how to override the default exposure and black point adjustment behavior in the histogram...

avidandrew:

[PR17838](https://github.com/darktable-org/darktable/pull/17838)

This documents how to modify the behavior when adjusting exposure/black point in the histogram, waveform, or rgb parade scopes.


elstoc:
IIRC there's also a bauhaus-style right-click option to edit these, which should probably also be documented.

avidandrew:
I made it a bit more clear that right-clicking is another way to fine-tune the adjustment

elstoc:
Ok thanks, looks good

PR #715, opened 2025-01-01
Changed the symbols for OR and AND to the Venn diagrams for union and intersection

avidandrew:

[PR17720](https://github.com/darktable-org/darktable/pull/17720)

This updates the screenshots for the color label filter to use the new Venn diagram symbols.


elstoc:
I think it would be better if the sizes of these images were closer to those currently on the website, which I made larger so they were clearer in the manual. IIRC I ended up playing around with the DPI and/or font sizes in `preferences>general` when I did these originally.

avidandrew:
I uploaded new, larger versions of these images. For reference, I set `GUI controls and text DPI` in `preferences>general` to 160 to get this new resolution

elstoc:
Thanks, could you modify the color labels image as well, and to include the same content as the previous image (show the entire widget, for consistency with the other images in the manual)

elstoc:
Nice one, thanks

PR #727, opened 2025-02-07
Added some words about vectorscope color harmony buttons only to be s…

DanielFeld2020:
…hown when in RYB color space


elstoc:
LGTM thanks!

PR #734, opened 2025-03-23
Document the find tag menu item in attached tags area

Phemisters:
Related to this PR: https://github.com/darktable-org/darktable/pull/16208

Add a doc note that there is a _find tag_ menu item in the tagging module for finding tags in the tag dictionary matching an attached tag.


Phemisters:
@elstoc this PR was created before any of my PRs had been merged and I think that is why there is still a workflow waiting to be approved.

elstoc:
I don't pay much attention to the workflow approval to be honest. The only reason I haven't looked at this yet is that I don't use tagging, so it requires more time than some of the others

PR #739, opened 2025-03-29
composite: describe 8bit limitation and workaround

ralfbrown:
Addresses https://github.com/darktable-org/darktable/issues/18573 by documenting the implementation limitations and how to avoid the resulting posterization or unexpected change in appearance of the overlaid image.


elstoc:
Thanks

PR #740, opened 2025-03-30
add note about handling non-hazy images

ralfbrown:
Related to discussion on https://github.com/darktable-org/darktable/issues/16746


jenshannoschwalm:
Diffuse/Sharpen would also be an obvious way.

PR #747, opened 2025-04-11
Add CIN and DPX to the list of graphicsmagic supported image formats

victoryforce:
See https://github.com/darktable-org/darktable/pull/18054 and https://github.com/darktable-org/darktable/pull/18280.


elstoc:
Thanks!

PR #749, opened 2025-04-12
Update collections sort / recent collections docs

Phemisters:
Related to: https://github.com/darktable-org/darktable/pull/16072

The sorting of collections is now done in a button (not a preference). There have also been some other minor changes to icons (collection _AND_ and _OR_ icons) and the way recent collections module is made visible.

I'm not sure of the best place to mention that you can now right-click in a side panel to see a list of utility modules and change their visibility - added to _overview_ but open to suggestions. Also, this might be taken as part of https://github.com/darktable-org/darktable/pull/17425, but that PR says it is a double click and that does not appear to be true.

Edit: Since I was removing the bit about the preference for altering the recent module's visibility I wanted to say something about this in this PR, but happy to remove the _overview_ change.


elstoc:
I'm fine with putting the note about hiding utility modules here. To complete the documentation on the new icons, it would be useful to also refresh the screenshot in "overview/user-interface/top-panel"

Phemisters:
>  To complete the documentation on the new icons, it would be useful to also refresh the screenshot in "overview/user-interface/top-panel"

Re-captured screenshot.

elstoc:
Thanks

PR #751, opened 2025-04-15
Add page on GIMP integration

Phemisters:
darktable PRs: 
* https://github.com/darktable-org/darktable/pull/16193
* https://github.com/darktable-org/darktable/pull/16232

Add a page describing GIMP integration. Although these are really  part of `program  invocation/darktable` I made a separate page as

1. the darktable page is already quite long
2. the GIMP parameters are not intended for users and are solely for GIMP's use

If it would be better on the main darktable parameters page I'll happily move it.

Incidentally, while experimenting with this I found a couple of bugs (https://github.com/darktable-org/darktable/issues/18608 and https://github.com/darktable-org/darktable/issues/18614) and the good news is they have both been fixed :smile:

### DIFF
--- a/content/special-topics/program-invocation/gimp-integration.md
+++ b/content/special-topics/program-invocation/gimp-integration.md
@@ -1,0 +1,39 @@
+---
+title: gimp integration
+id: gimp-integration
+weight: 85
+draft: false
+author: "people"
+---
+
+GIMP can use darktable to open a RAW file if the darktable executable is on the PATH. If the GIMP user opens a RAW file, then GIMP invokes darktable with some special command line parameters to get a temporary file for GIMP to work with. It uses these command line parameters:
+
+```
+darktable --gimp [version]
+                 [file <path>]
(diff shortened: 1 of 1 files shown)


jenshannoschwalm:
You might want to mention, that gimp will call darktable using it's edits done before in darkroom but won't keep any edits done while working as a gimp plugin.

Phemisters:
> gimp will call darktable using it's edits done before in darkroom but won't keep any edits done while working as a gimp plugin.

@jenshannoschwalm - Just to sure I understand, do you mean that 

1. when darktable is invoked by gimp, if the image is already in the  darktable database, then any edits previously done in darktable will be part of the temp image exported by darktable
2. any edits during that invocation of darktable (prior to the export of the temp file) will not be saved to the darktable database.

jenshannoschwalm:
Yes .

Phemisters:
As suggested, added note about prior edits being included, but new edits not being saved.

PR #757, opened 2025-04-17
Color picker and distorting modules doc change

Phemisters:
darktable PR: https://github.com/darktable-org/darktable/pull/16534

I have copied the text from the [release description](https://www.darktable.org/2024/06/darktable-4.8.0-released/) since (after trying it out) I couldn't come up with a better way of describing this change. In fact, it may be that we don't need to add the extra text at all, as I suppose it now works the way people might have assumed it worked anyway.

I also fixed what seem to be broken references to the icon.


elstoc:
The bit about the icon was supposed to be a link to the section of the manual that describes pickers (in general), rather than an icon (I assume that link does work). Also I think that extra sentence poses more questions (to the user) than it answers.

So I don't think we need this PR, though I'm open to discussion if you disagree.

Phemisters:
> The bit about the icon was supposed to be a link to the section of the manual that describes pickers (in general), rather than an icon (I assume that link does work). 

For me the existing link results in a 404 response, and it is not clear to me what the intention is if its not the icon.

> 
> So I don't think we need this PR, though I'm open to discussion if you disagree.

I'm happy to close the PR.

elstoc:
> For me the existing link results in a 404 response, and it is not clear to me what the intention is if its not the icon.

Yes you're right, the link is broken. Fixed in ff09e608527f73f567724c352a1199d599774fe3

PR #758, opened 2025-04-17
Update import page in doc

Phemisters:
darktable PR: https://github.com/darktable-org/darktable/pull/17542

In the import dialog there is now a button (`-`) for removing custom places instead of right clicking.

Also updated a couple of descriptions from the copy & import dialog.

Still not very sure about the correct size for capturing images (maybe *too* big now) - I hope to get better eventually!


elstoc:
Thanks

PR #759, opened 2025-04-18
Update doc for default tone-mapper workflow

Phemisters:
darktable PR: https://github.com/darktable-org/darktable/pull/18547

Update the default in preferences for the workflow - I moved sigmoid up as it is the new default and that also makes the description order correspond to the order of the combo-box.

While doing this I noticed the white balance now has "as shot to reference" and for me that seems to be what it defaults to. That is why I updated the sentence about white balance, as I think it applies in the same way as for "camera reference". You may prefer we wait to do that change when darktable PR [15602](https://github.com/darktable-org/darktable/pull/15602#issuecomment-2712865644) is done.


elstoc:
Thanks. I will hold off merging this for now as I suspect we might need to make some other tweaks elsewhere in the doc to account for this. For example, while I'm not currently interested in changing the _recommendations_ in the overview/workflow section (which suggests using filmic to process), it might still need some wording tweaks to account for the new default.

Phemisters:
> we might need to make some other tweaks 

I'm not sure if you have a plan for this. I have had a go at some small updates for this, but happy to revert those or indeed close this PR until someone has the time and knowledge for a more thorough update.

PR #762, opened 2025-04-26
Update scale_pixels module description

da-phil:
The description was missing a little bit more context and the fact that the module has a parameter.

This doc udpate is not related to a PR in the dt repo, it's just an independent update.


elstoc:
Thanks

PR #770, opened 2025-06-11
document the interchangebility of the rating and color label

UweOhse:
This is a pure documentation addendum.

I wasted a number of hours yesterday trying to figure out why my color labels didn't show up in another application, and wanted to spare others the same experience.

### DIFF
--- a/content/lighttable/digital-asset-management/star-color.md
+++ b/content/lighttable/digital-asset-management/star-color.md
@@ -24,6 +24,8 @@ To rate multiple images at once, select those images in the lighttable or filmst
 
 You can filter images by star rating in the [top panel](../../overview/user-interface/top-panel.md).
 
+The rating of an image is written to the xmp:Rating metadata field, and is compatible with other applications.
+
 # color labels
 
 Color labels are another way to classify images, and can be used as an alternative to star ratings or to work alongside them. Each image can carry any combination of one or more color labels (red, yellow, green, blue, or purple).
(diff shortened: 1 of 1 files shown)


elstoc:
I agree that this is useful information to have in the user manual. A couple of thoughts for consideration.

1. You talk about the information being "stored in a metadata field" but it's not clear what you mean by that. Do you mean "within the sidecar file" or "within an exported image" or both? The information is available in both but I'm guessing you mean the exported image?
2. You also talk about compatibility with other applications. This is a difficult one, and I'd be reluctant to pass comment on compatibility with anything outside of darktable, especially since it's not clear what we might mean by compatibility. For example, both of these fields can be queried by a tool that reads image metadata, and both are shown (albeit the color labels is a number so hard to interpret) when I view the image in geeqie. I could probably work out how to access the color label information in _some_ other applications if I really wanted to. Perhaps we could word this better without speculating on other applications' functionality.
3. Finally, the rating is exported to Xmp.xmp.Rating, but from what I can see it's also exported to Exif.Image.Rating which is possibly a better field to reference.

Phemisters:
@UweOhse - this is another request for a response to the comments on this PR. If you are not interested in updating the PR, it will probably be closed.

Phemisters:
Despite several comments there has been no response. Interchangeability with other applications is hard to maintain without access to the other applications. I'm closing this for now, it can re-opened if someone wants to handle this.

PR #780, opened 2025-09-17
Replace use of camera manufacturer with maker

Phemisters:
Related darktabke issue: https://github.com/darktable-org/darktable/pull/18897

Replace uses of _camera manufacturer_ with _camera maker_. I'm not sure if the sentences now sometimes read a little oddly, but this makes it more consistent with darktable itself.

I have not changed other uses of manufacturer (e.g. printer or GPU manufacturer).


elstoc:
LGTM, thanks!

PR #786, opened 2025-11-02
Clarity and structure improvements in the shortcuts documentations

masterpiga:
The documentation about shortcuts was confusing to me. After a few exchanges on pixls.us, I think that I understand things better. Armed with this better understanding, I thought that I was in a good position to try to improve it.

The content is by and large the same, I have just moved things around a bit to keep the information flow more consistent and to add a bit of structure.

In terms of content changes, the only notable one is that I replaced the distinction between **simple** vs. **extended** shortcuts, which seemed a bit arbitrary and that I found confusing. Instead, I used **discrete** vs. **continuous**, as these shortcuts are associated with **discrete** (e.g., click) vs. **continuos** (e.g., change value) UI actions. This also shifts the focus from the type of shortcut that one is defining to the kind of action that one wants to trigger, which IMHO is more logical and hence more understandable and memorable.

I also added a few more concrete examples, which were lacking, and whenever possible aded captions to the paragraph to make it easier to find what one is looking for.

**Edit**: Following some review iterations, the remapped terminology is now **simple --> discrete** and **extended --> directional**.

### DIFF
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -148,7 +148,7 @@ For example, you can assign the shortcut `e + scroll` to change the value (the _
 
 # assigning shortcuts to actions
 
-There are two primary methods of assigning shortcuts to actions: [visual shortcut mapping](#visual-shortcut-mapping) (recommended) and the [shortcut mapping screen](#shortcut-mapping-screen). The former makes it easier to remap multiple shortcuts at once, while the latter allows for finer control and more flexibility. Both options are detailed below.
+There are two primary methods of assigning shortcuts to actions: [visual shortcut mapping](#visual-shortcut-mapping) (recommended) and the [shortcut mapping screen](#shortcut-mapping-screen). The former makes it easier to remap multiple shortcuts at once, while the latter allows for finer control and more flexibility.
 
 ## visual shortcut mapping
 
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -39,7 +39,7 @@ Similarly, to control a slider with a shortcut you can:
 The system is very flexible and powerful, and this is just scratching the surface.
 Keep on reading to understand it better and learn how to make the most of it.
 
-# types of actions and shortcuts
+# types of shortcuts and actions
 
 There are two types of actions:
 
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -47,9 +47,13 @@ If you have an external controller, you can trigger discrete actions by pressing
 # anatomy of a shortcut
 
 The simplest shortcut consists of just one key press (e.g., `E`).
+
 The same key can be repeated up to three times, so also `E + E` and `E + E + E` are valid, and distinct, shortcuts.
+
 You can further extend a shortcut with up to three clicks of different mouse buttons. So, `E + E + left-click` or `E + right-click + left-click` are also valid shortcuts.
+
 Finally, if your fingers are very nimble and you really need to go there, you can repeat the last click up to three times. So, also `E + left-click + middle-click + middle-click` and `E + E + middle-click + left-click + right-click + right-click + right-click` are valid shortcuts.
+
 Hence, the most complex shortcut can consist of 8 key/button presses:
 * Three key presses of the same `key`, followed by
 * Three presses of the three mouse buttons, followed by
@@ -66,13 +70,13 @@ If you are a MacOs user, your shortcuts will use `Cmd` instead of `Ctrl` and `Op
 
 If you are defining a continuous shortcut, then the movement part of the shortcut must be executed while the last key, mouse or controller button is held down.
 
-For example, the shortcut `E + E + pan`, can be activated by pressing `E` twice, holding down `E` on the second press and moving the mouse horizontally. While you are holding down `E`, moving the mouse horizontally will adjust the value associated with the shortcut's action. As you release the `E` key, moving the mouse horizontally will just move the mouse on the screen.
+For example, the shortcut `E + E + pan`, can be activated by pressing `E` twice, holding down `E` on the second press and moving the mouse horizontally and moving the mouse horizontally will the key is pressed.
 
-**Long keypresses.** 
-The last repetition of a key in a shortcut can be either a _short_ (i.e., normal) or _long_ keypress, defined as holding down the key for a bit longer than the duration of a double click. Hence `E + E` and `E + E(long)` are two distinct shortcuts that can be assigned to different actions. The associated action triggers when the key is released, which entails that a shortcut ending with a long press cannot be used for a continuos shortcut.
+**Short and long keypresses.** 
+By default, all keypresses in a shortcut are _short_, i.e., the key is pressed and immediately released. However, the last repetition of a key in a shortcut can also be a _long_ keypress, defined as holding down the key for a bit longer than the duration of a double click. Hence `E + E` and `E + E(long)` are two distinct shortcuts that can be assigned to different actions. The associated action triggers when the key is released, which entails that a shortcut ending with a long press cannot be used for a continuos shortcut.
 
 **Triggering multiple shortcuts at once.** Note that `E + A` is not a valid shortcut, and darktable will interpret it as a two different shortcuts: `E` followed by `A`. This is by design, as the system allows one to trigger multiple continuous shortcuts at once.
-For example, if both `E + scroll` and `A + scroll` are mapped to a slider (or if you have fallbacks enabled), then scrolling with the mouse while holding down both `E` and `A` will move both sliders. If you have a series of keys assigned to nodes in a curve (e.g, tone equalizer) this allows you to move multiple nodes in parallel.
+For example, if both `E + mouse-scroll` and `A + mouse-scroll` are mapped to a slider (or if you have fallbacks enabled), then scrolling with the mouse while holding down both `E` and `A` will move both sliders. If you have a series of keys assigned to nodes in a curve (e.g, tone equalizer) this allows you to move multiple nodes in parallel.
 
 **Shortcuts must be unique within a view.**
 A single action may have multiple shortcuts but a single shortcut can only be linked to one action in a given darktable view -- you can't chain actions together except by applying a preset or style. You can, however, set up a single shortcut that does one thing in the lighttable view, say, and another in the darkroom view.
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -64,7 +64,7 @@ Finally, if your fingers are very nimble and you really need to go there, you ca
 Hence, the most complex shortcut can consist of 8 key/button presses:
 * Three key presses of the same `key`, followed by
 * Three presses of the three mouse buttons, followed by
-* 2 more repetions of the last click.
+* 2 more repetitions of the last click.
 
 Your shortcut can include one or more modifiers (`Shift`, `Ctrl` and `Alt`). In this case, the modifier(s) have to be held down while executing the remainder of the shortcut. So, `Ctrl + E + E` means holding down `Ctrl` while pressing `E` twice in a rapid sequence.
 
@@ -79,11 +79,12 @@ If you are defining a continuous shortcut, then the movement part of the shortcu
 
 For example, the shortcut `E + E + pan`, can be activated by pressing `E` twice, holding down `E` on the second press and moving the mouse horizontally and moving the mouse horizontally will the key is pressed.
 
-**Short and long keypresses.** 
-By default, all keypresses in a shortcut are _short_, i.e., the key is pressed and immediately released. However, the last repetition of a key in a shortcut can also be a _long_ keypress, defined as holding down the key for a bit longer than the duration of a double click. Hence `E + E` and `E + E(long)` are two distinct shortcuts that can be assigned to different actions. The associated action triggers when the key is released, which entails that a shortcut ending with a long press cannot be used for a continuos shortcut.
+**Short and long key presses.** 
+By default, all key presses in a shortcut are _short_, i.e., the key is pressed and immediately released. However, the last repetition of a key in a shortcut can also be a _long_ keypress, defined as holding down the key for a bit longer than the duration of a double click. Hence `E + E` and `E + E(long)` are two distinct shortcuts that can be assigned to different actions. The associated action triggers when the key is released, which entails that a shortcut ending with a long press cannot be used for a continuos shortcut.
 
-**Triggering multiple shortcuts at once.** Note that `E + A` is not a valid shortcut, and darktable will interpret it as a two different shortcuts: `E` followed by `A`. This is by design, as the system allows one to trigger multiple continuous shortcuts at once.
-For example, if both `E + mouse-scroll` and `A + mouse-scroll` are mapped to a slider (or if you have fallbacks enabled), then scrolling with the mouse while holding down both `E` and `A` will move both sliders. If you have a series of keys assigned to nodes in a curve (e.g, tone equalizer) this allows you to move multiple nodes in parallel.
+**Triggering multiple shortcuts at once.** 
+If both `E + mouse-scroll` and `F + mouse-scroll` are mapped to a slider (or if you have fallbacks enabled), then scrolling the mouse wheel while holding down both `E` and `F` will move both sliders.
+Similarly, you could map, say, `E + E + mouse-scroll` and `F + F + mouse-scroll` to two nodes in a graph (say, `yellow` and `green` in color equalizer). You can then use `E + E(hold) + F(hold) + mouse-scroll` to move both sliders at once. This works because the repeat key counter is per-shortcut, not per key, and it is reset only after the current shortcut ends (i.e., when the repeated key is released). Thus it is possible to trigger as many actions at once as your fingers can handle, provided that all the shortcuts involved have the same number of key presses (i.e., single, double or triple presses).
 
 **Shortcuts must be unique within a view.**
 A single action may have multiple shortcuts but a single shortcut can only be linked to one action in a given darktable view -- you can't chain actions together except by applying a preset or style. You can, however, set up a single shortcut that does one thing in the lighttable view, say, and another in the darkroom view.
@@ -235,7 +236,7 @@ _enable_
 : Acts as a _toggle_ that switches the module on and off.
 
 _focus_
-: Acts as a _toggle_ that focuses or defocuses the module. This is useful for modules such as [_crop_](../module-reference/processing-modules/crop.md) or [_tone equalizer_](../module-reference/processing-modules/tone-equalizer.md), whose on-screen controls are only activated when those modules have focus. For _crop_, changes are saved only when the module loses focus.
+: Acts as a _toggle_ that focuses or de-focuses the module. This is useful for modules such as [_crop_](../module-reference/processing-modules/crop.md) or [_tone equalizer_](../module-reference/processing-modules/tone-equalizer.md), whose on-screen controls are only activated when those modules have focus. For _crop_, changes are saved only when the module loses focus.
 
 _instance_
 : Allows you to select actions from the [multiple-instance](../darkroom/processing-modules/multiple-instances.md) menu (e.g. move up/down, create new instance). The discrete action associated to the _instance_ element displays a list of the available options for selection; a continuous action is also available and will move the _preferred module instance_ (see below) up and down the pixelpipe.
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -5,7 +5,7 @@ weight: 120
 draft: false
 ---
 
-_Shortcuts_ are a way to control darktable using your keyboard, mouse or trackpad and other input devices. They allow you to perform _actions_ without interacting directly with a UI element.
+_Shortcuts_ are a way to control darktable using your keyboard, mouse, trackpad, or other input devices. They allow you to perform _actions_ without interacting directly with a UI element.
 
 An _action_ is usually an operation that you might undertake using darktable's point-and-click user interface.
 See [common actions](#common-actions) for a list of the most common ones. For example:
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -60,7 +60,7 @@ Your shortcut can include one or more modifiers (`Shift`, `Ctrl` and `Alt`). In 
 ---
 
 **Note:**
-If you are a MacOs user, your shortcuts will use `Cmd` instead of `Ctrl` and `Option` instead of `Alt`
+If you are a MacOs user, your shortcuts will use `Cmd` instead of `Ctrl` and `Option` instead of `Alt`.
 
 ---
 
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -181,7 +181,7 @@ left-click on a mappable widget to open the shortcut mapping screen for that wid
 define a second shortcut against the same widget. If you attempt to reallocate an existing shortcut to a new action, you will be notified of the conflict and prompted to replace the existing shortcut.
 
 **To change the default speed of a slider:**
-scroll with your mouse wheel while in visual mapping mode (without pressing any other buttons/keys) when hovering over the slider -- scroll up to increase and down to decrease. When you leave mapping mode, normal mouse scrolls over that slider will change its value with the adjusted speed.
+Scroll with your mouse wheel while in visual mapping mode (without pressing any other buttons/keys) when hovering over the slider -- scroll up to increase and down to decrease. When you leave mapping mode, normal mouse scrolls over that slider will change its value with the adjusted speed.
 
 ## shortcut mapping screen
 
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -62,7 +62,7 @@ For example:
 ---
 
 **Note:**
-By construction, a _directional shortcut_ is sufficiently expressive to trigger a _discrete action_. The opposite is not true. However, it is possible to use a combination of two _discrete shortcuts_ to control a _directional action_. For example, you can use two different shortcuts to trigger the `up` and `down` effects of a slider (see [anatomy of an action](#anatomy-of-an-action)).
+A _directional shortcut_ is sufficient to trigger a _discrete action_. The opposite is not true. However, it is possible to use a combination of two _discrete shortcuts_ to control a _directional action_. For example, you can use two different shortcuts to trigger the `up` and `down` effects of a slider (see [anatomy of an action](#anatomy-of-an-action)).
 
 ---
 
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -171,7 +171,7 @@ Press a key combination while hovering over a mappable widget. A _default action
 You can assign as many shortcuts as you like in a single mapping session and then exit mapping mode when you are finished by clicking the ![visual mapping button](./shortcuts/visual-mapping-button.png#icon) icon again or right-clicking anywhere on the screen.
 
 **To explore already defined shortcuts:**
-left-click on a mappable widget to open the shortcut mapping screen for that widget (see below). Left-click anywhere else on the screen to open the shortcut mapping screen, expanded (where possible) based on the part of the screen you have clicked on. This screen can be used to alter the action assigned to a shortcut and to configure shortcuts for non-visual actions. Entering the shortcut mapping screen exits visual shortcut mapping mode.
+Left-click on a mappable widget to open the shortcut mapping screen for that widget (see below). Left-click anywhere else on the screen to open the shortcut mapping screen, expanded (where possible) based on the part of the screen you have clicked on. This screen can be used to alter the action assigned to a shortcut and to configure shortcuts for non-visual actions. Entering the shortcut mapping screen exits visual shortcut mapping mode.
 
 **To update existing shortcuts:**
 define a second shortcut against the same widget. If you attempt to reallocate an existing shortcut to a new action, you will be notified of the conflict and prompted to replace the existing shortcut.
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -306,3 +306,9 @@ To see a list of _all_ of the default fallbacks, click the "enable fallbacks" ch
 Fallbacks are only applied if no other shortcut using the resulting combination has been explicitly created. In the above example, if you were to explicitly assign `Ctrl + R + left-click` to another action, the "enable/disable module" fallback would be ignored.
 
 Some fallback actions are defined using modifier keys (usually `Ctrl+` and `Shift+`). In this case you must define an initial shortcut without such a modifier in order to be able to use these fallbacks. For example, if you assign `Ctrl + R` to an action, you cannot use a `Ctrl+` fallback. Some default fallbacks of this type are provided for the _value_ element and for horizontal/vertical movements in the (zoomed) central area -- in this case, `Shift+` increases the speed to 10.0 and `Ctrl+` decreases the speed to 0.1.
+
+# implementation details
+
+This section covers some technical aspects that can be interesting for power users.
+
+**Continuous actions:** Continuous actions are convenience abstractions to make assigning shortcuts faster, but are actually implemented as two discrete actions that target the appropriate effects (e.g., `up` and `down`) of a widget. For example, assigning `E + pan` to a slider assignes `E + left` to the `down` action and `E + right` to the `up` action.
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -192,7 +192,7 @@ The top panel of the shortcut mapping screen shows a list of available UI widget
 ---
 
 **Note:**
-You can **search** the top and bottom panels using the text entry boxes at the bottom of the screen, and use the **up/down** arrow keys to navigate between matches. 
+You can search the top and bottom panels using the text entry boxes at the bottom of the screen, and use the up/down arrow keys to navigate between matches. 
 
 ---
 
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -40,6 +40,13 @@ For example:
 - The key `E` can be used to focus a module, or to toggle it on/off
 - The key `E`, combined with up/down movements or your pointing device, can be used to control the value of a slider
 
+---
+
+**Note:**
+By construction, a _continuous shortcut_ is sufficiently expressive to trigger a _discrete action_. The opposite is not true. However, it is possible to use a combination of two _discrete shortcuts_ to control a _continuous action_. For example, you can use two different shortcuts to trigger the `up` and `down` effects of a slider (see [anatomy of an action](#anatomy-of-an-action) below).
+
+---
+
 If you are using keyboard and mouse, **all shortcuts must start with one or more key presses**, as mouse actions in isolation are used to navigate and interact with the UI.
 
 If you have an external controller, you can trigger discrete actions by pressing one or more buttons on the controller. Continuous actions can use a combination of buttons and a knob/joystick movement, or just knob/joystick movement.
@@ -310,9 +317,3 @@ To see a list of _all_ of the default fallbacks, click the "enable fallbacks" ch
 Fallbacks are only applied if no other shortcut using the resulting combination has been explicitly created. In the above example, if you were to explicitly assign `Ctrl + R + left-click` to another action, the "enable/disable module" fallback would be ignored.
 
 Some fallback actions are defined using modifier keys (usually `Ctrl+` and `Shift+`). In this case you must define an initial shortcut without such a modifier in order to be able to use these fallbacks. For example, if you assign `Ctrl + R` to an action, you cannot use a `Ctrl+` fallback. Some default fallbacks of this type are provided for the _value_ element and for horizontal/vertical movements in the (zoomed) central area -- in this case, `Shift+` increases the speed to 10.0 and `Ctrl+` decreases the speed to 0.1.
-
-# implementation details
-
-This section covers some technical aspects that can be interesting for power users.
-
-**Continuous actions:** Continuous actions are convenience abstractions to make assigning shortcuts faster, but are actually implemented as two discrete actions that target the appropriate effects (e.g., `up` and `down`) of a widget. For example, assigning `E + pan` to a slider assignes `E + left` to the `down` action and `E + right` to the `up` action.
--- a/content/preferences-settings/shortcuts.md
+++ b/content/preferences-settings/shortcuts.md
@@ -5,10 +5,10 @@ weight: 120
 draft: false
 ---
 
-_Shortcuts_ are a way to control darktable using your keyboard, pointing device or external controller. They allow you to perform _actions_ without interacting directly with a UI element.
+_Shortcuts_ are a way to control darktable using your keyboard, mouse or trackpad and other input devices. They allow you to perform _actions_ without interacting directly with a UI element.
 
-An _action_ is usually (but not always) an operation that you might undertake using darktable's point-and-click user interface.
-See [common actions](#common-actions) below for a list of the most common ones. For example:
+An _action_ is usually an operation that you might undertake using darktable's point-and-click user interface.
+See [common actions](#common-actions) for a list of the most common ones. For example:
 
 -   Increase, decrease or reset sliders
 -   Scroll through dropdown lists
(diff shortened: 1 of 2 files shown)


masterpiga:
Apologies, done.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,97 @@ weight: 120
...
 
-For external devices you do not need to start with a _simple_ shortcut -- you can directly assign a control knob or joystick to an action -- though this will significantly reduce the flexibility of such devices.
+**Shortcuts must be unique within a view.**
+A single action may have multiple shortcuts but a single shortcut can only be linked to one action in a given darktable view -- you can't chain actions together except by applying a preset or style. You can, however, set up a single shortcut that does one thing in the lighttable view, say, and another in the darkroom view.
 
-Long button and key presses cannot be extended, as the length of the click/press is timed using the release of the final button/key -- such shortcuts must be terminated with the raising of the final button/key.
+**Additional modifiers.** As mentioned above, the only valid modifiers are the Shift, Ctrl/Cmd and Alt/Option keys on the keyboard. You can define additional keys (or device buttons) as modifiers by assigning keys/buttons to the "global/modifier" action. However, these will merely function as extra Ctrl, Alt or Shift keys -- you cannot create "new" modifiers.
```
dterrahe: Are Cmd and Option for Apple users mentioned everywhere in the docs where Ctrl and Alt are used? Or should/is this be something that the introduction mentions once. It could be somewhat confusing for everyone else.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,97 @@ weight: 120
...
+If you are using keyboard and mouse, **all shortcuts must start with one or more key presses**, as mouse actions in isolation are used to navigate and interact with the UI.
 
-### creating additional modifiers
+If you have an external controller, you can trigger discrete actions by pressing one or more buttons on the controller. Continuous actions can use a combination of buttons and a knob/joystick movement, or just knob/joystick movement.
 
-The only valid modifiers are the Shift, Ctrl and Alt keys on the keyboard. You can define additional keys (or device buttons) as modifiers by assigning keys/buttons to the "global/modifier" action. However, these will merely function as extra Ctrl, Alt or Shift keys -- you cannot create "new" modifiers.
+# anatomy of a shortcut
 
-## extending simple shortcuts with movement
+Shortcuts can consist of up to three key/button presses **of the same key/button** (in a quick sequence).
```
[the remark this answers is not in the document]
masterpiga: Thanks, very informative. Updated accordingly.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,97 @@ weight: 120
...
+- _Continuous actions_ need also a direction (and, in the case of sliders, a magnitude). Hence, the corresponding _continuous shortcuts_ will also incorporate a mouse movement
 
-A simple shortcut must be initiated as above, but can include:
+For example:
 
--   One or more modifier keys (Shift, Ctrl, Alt), held down while executing the remainder of the shortcut
--   Up to three key presses, the last one of which may be a long-press (defined as a key-press longer than your system's double-click duration)
--   Similarly, up to three device-button presses or mouse-button clicks, the last of which may be long
+- The key `e` can be used to focus a module, or to toggle it on/off
+- The key `e`, combined with up/down movements or your pointing device, can be used to control the value of a slider
```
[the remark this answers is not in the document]
masterpiga: I mention long presses on line 62, explaining that if the shortcut includes a movement the movement must be performed as the last key is held down.

My understanding is that long presses only make sense in this context, but I suspect that you will tell me that I am wrong :)

content/preferences-settings/shortcuts.md
```diff
@@ -46,21 +46,35 @@ If you have an external controller, you can trigger discrete actions by pressing
 
 # anatomy of a shortcut
 
-Shortcuts can consist of up to three key/button presses **of the same key/button** (in a quick sequence).
-So, for example, `E`, `E + E` and `E + E + E` are all valid shortcuts to trigger discrete actions, as well as `E + left-click`, `E + right-click` or `E + E + left-click`.
+The simplest shortcut consists of just one key press (e.g., `E`).
+The same key can be repeated up to three times, so also `E + E` and `E + E + E` are valid, and distinct, shortcuts.
+You can further extend a shortcut with up to three clicks of the same mouse button. So, `E + E + left-click` or `E + right-click + right-click` are also valid shortcuts.
+Finally, if your fingers are very nimble and you really need to go there, you can add either or both of the mouse buttons that you haven't incorporated yet. So, also `E + left-click + middle-click` and `E + E + right-click + right-click + right-click + left-click + middle-click` are valid shortcuts.
```
dterrahe: Aaah! Now with more testing to remind me, the "triple click" of the mouse button has to be the _final_ button. So it would have to be `E+E+E+right+middle+left+left+left`.

content/preferences-settings/shortcuts.md
```diff
@@ -46,21 +46,35 @@ If you have an external controller, you can trigger discrete actions by pressing
...
+Hence, the most complex shortcut can consist of 8 key/button presses:
+* Three key presses of the same `key`, followed by
+* Three presses of any mouse button, followed by
+* 1 press of each remaining mouse button.
 
-On the contrary, `E + A` is not a valid shortcut, and darktable will interpret it as a two different shortcuts: `E` followed by `A`.
+Your shortcut can include one or more modifiers (`Shift`, `Ctrl` and `Alt`). In this case, the modifier(s) have to be held down while executing the remainder of the shortcut. So, `Ctrl + E + E` means holding down `Ctrl` while pressing `E` twice in a rapid sequence.
```
dterrahe: I'm not suggesting any wording here; just trying to help improve _your_ mental modal and then you can decide if it is worth explaining to the wider public.

You _don't_ have to press the modifiers before pressing the shortcut keys. You can add them later. So for example if triple-press A is assigned to a slider (and lets assume you have fallbacks enabled). You can do A+A+A hold, then scroll for "normal sized" adjustments. Then (while still holding the A) add Ctrl for some final fine-tuning. Personally I prefer mouse  moves, so I first scroll and then make some final left-right moves at *.1.

content/preferences-settings/shortcuts.md
masterpiga: Oh well, this makes much more sense :) Fixed.

content/preferences-settings/shortcuts.md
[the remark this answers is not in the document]
masterpiga: Aaaaaaall right, I had gotten this completely wrong. Reintegrated. I played with it a bit, and I must confess that I find the long press very difficult to get right. If you press too long it is just ignored, and since it's not easy to get the timing right it can be a bit confusing.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,97 @@ weight: 120
...
 
-The only valid modifiers are the Shift, Ctrl and Alt keys on the keyboard. You can define additional keys (or device buttons) as modifiers by assigning keys/buttons to the "global/modifier" action. However, these will merely function as extra Ctrl, Alt or Shift keys -- you cannot create "new" modifiers.
+# anatomy of a shortcut
 
-## extending simple shortcuts with movement
+Shortcuts can consist of up to three key/button presses **of the same key/button** (in a quick sequence).
+So, for example, `E`, `E + E` and `E + E + E` are all valid shortcuts to trigger discrete actions, as well as `E + left-click`, `E + right-click` or `E + E + left-click`.
 
-For certain actions you can choose to _extend_ a simple shortcut using mouse/device movement. For example you might hold Ctrl+X while scrolling with your mouse to change the value of a slider. The following can be used to extend a simple shortcut:
+On the contrary, `E + A` is not a valid shortcut, and darktable will interpret it as a two different shortcuts: `E` followed by `A`.
```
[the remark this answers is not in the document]
dterrahe:
```
F1;double=iop/colorequal/brightness/red
F1;triple=iop/toneequal/blacks
F2;double=iop/colorequal/brightness/orange
F2;triple=iop/toneequal/deep shadows
F3;double=iop/colorequal/brightness/yellow
F3;triple=iop/toneequal/shadows
F4;double=iop/colorequal/brightness/green
F4;triple=iop/toneequal/light shadows
F5;double=iop/colorequal/brightness/cyan
F5;triple=iop/toneequal/mid-tones
F6;double=iop/colorequal/brightness/blue
F6;triple=iop/toneequal/dark highlights
F7;double=iop/colorequal/brightness/lavender
F7;triple=iop/toneequal/highlights
F8;double=iop/colorequal/brightness/magenta
F8;triple=iop/toneequal/whites
```
You're probably overthinking this. Merge the above and `F1+F1+F2+F3+scroll` will move three color equalizer nodes (if you have fallbacks enabled) and `F3+F3+F3+F2+F1+scroll` will move three tone equalizer nodes.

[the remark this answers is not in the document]
masterpiga: Ok, thanks for the snippet, I tried it and yes, I can move the three nodes together. I still do not understand the logic, though.

As exemplified above, `iop/colorequal/brightness/orange` is triggered by `F2,double`, which I parse as "double press of F2".

In the key combination `F1 + F1 + F2` there is no double press of `F2`, so why should `iop/colorequal/brightness/orange` trigger?

I have seen that it does. The question is why? It shouldn't according to (how I parse) the specification of the shortcut.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,97 @@ weight: 120
...
-You cannot initiate a shortcut by moving your mouse, or by pressing the left, right or middle mouse buttons, as these actions are used to interact with darktable's UI.
+_Continuous actions_
+: E.g., adjusting the value of a slider or scrolling through a list of presets
 
-## simple shortcuts
+As the two action types are inherently different, so are the shortcuts you can use to trigger them:
 
-A shortcut that only includes button and/or key presses (and not mouse/device movements) is referred to as a _simple_ shortcut.
+- _Discrete actions_ can be triggered by _discrete shortcuts_ consisting only of key presses and optionally mouse clicks
+- _Continuous actions_ need also a direction (and, in the case of sliders, a magnitude). Hence, the corresponding _continuous shortcuts_ will also incorporate a mouse movement
```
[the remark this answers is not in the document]
masterpiga: Ok, in the light of this I removed the "implementation details" section and added line explaining that continuous shortcuts can also be used to trigger discrete actions.

content/preferences-settings/shortcuts.md
dterrahe: Shortcuts don't get triggered when you _press_ a key (or button); they get triggered when you _release_ a key (ok, with the exception of _hold_ shortcuts...). Modifiers don't trigger shortcuts, they only "modify" them. So if you press e, then Ctrl then release e, you should get the Ctrl+e shortcut. If you do it _fast_ enough, otherwise it will be a long press or ignored. 

So it has nothing to do with fallbacks but it is more relevant for what you call continuous shortcuts. But also if you combine keypresses with mouse clicks. Ctrl+e+left click you can add the Ctrl _after_ pressing (and holding) the e (_before_ or even _during_ the click).

Disclaimer not behind keyboard so not testing/verifying right now.

elstoc:
For me, the problem with big restructure PRs like this is that they're essentially unreviewable using github (or really at all TBH). Yes we can make small comments on individual paragraphs/terminology/use of markdown, but it's really almost impossible to have a reasonable discussion about the broader-brush changes because there's so much of it.

Such discussions almost always devolve into a dozen threads of inconclusive back-and-forth without addressing the whole. The only options from my point of view are to wait until it's stable, merge it as-is and copy-edit (which might turn into another rewrite) or reject it entirely (which would be a shame as I'm sure there are some good ideas here).

I spent quite a long time structuring and rewriting this section to try and make it flow well (it's complex functionality and explaining it well and consistently is hard). That's not to say it can't be improved but I really don't have the time it would take to give such a large PR the attention it deserves.

I now have much less time to devote to dtdocs than I used to and I would prefer to spend my time documenting new features. In my opinion this PR would be much better (for my sanity as much as anything) as a set of small issues raised and small PRs to resolve them one at a time.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,126 @@ weight: 120
...
+See [common actions](#common-actions) below for a list of the most common ones. For example:
 
-A _shortcut_ is a combination of key or button presses and/or mouse or device movements that performs an _action_ in darktable.
+-   Increase, decrease or reset sliders
+-   Scroll through dropdown lists
+-   Enable, expand or focus modules
+-   Click buttons
+-   Switch between views
 
-The recommended way to assign shortcuts to visual elements is the [visual shortcut mapping](#visual-shortcut-mapping) mode.
```
paperdigits: Is this no longer the recommended way to assign shortcuts? This should be mentioned early on if it is still the recommended way.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,126 @@ weight: 120
...
+* Three presses of the three mouse buttons, followed by
+* 2 more repetitions of the last click.
 
-Long button and key presses cannot be extended, as the length of the click/press is timed using the release of the final button/key -- such shortcuts must be terminated with the raising of the final button/key.
+Your shortcut can include one or more modifiers (`Shift`, `Ctrl` and `Alt`). In this case, for a continuous shortcut, both the last key and the modifier must be pressed down when you execute the movement. For example, `Ctrl + E + E + pan` means that while moving the mouse left and/or right both  `E` and `Ctrl` must be pressed down.
 
 ---
 
-**Note:** You may need to switch off the "disable touchpad while typing" setting if you want to use extended shortcuts with a laptop touchpad.
+**Note:**
```
paperdigits: This note should probably go in the introductory section, if we include it at all.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,126 @@ weight: 120
...
+**Short and long key presses.** 
+By default, all key presses in a shortcut are _short_, i.e., the key is pressed and immediately released. However, the last repetition of a key in a shortcut can also be a _long_ key press, defined as holding down the key for a bit longer than the duration of a double click. Hence `E + E` and `E + E(long)` are two distinct shortcuts that can be assigned to different actions. The associated action triggers when the key is released, which entails that a shortcut ending with a long press cannot be used for a continuos shortcut.
 
--   Increase, decrease or reset sliders
--   Scroll through dropdown lists
--   Enable, expand or focus modules
--   Click buttons
--   Switch between views
+**Triggering multiple shortcuts at once.** 
+If both `E + mouse-scroll` and `F + mouse-scroll` are mapped to a slider (or if you have fallbacks enabled), then scrolling the mouse wheel while holding down both `E` and `F` will move both sliders.
```
paperdigits: It would be nice to refer to the short cuts the same way as the preferences panel does, e.g. it doesn't call it "mouse scroll" just "scroll".

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,126 @@ weight: 120
...
+
+**Shortcuts must be unique within a view.**
+A single action may have multiple shortcuts but a single shortcut can only be linked to one action in a given darktable view -- you can't chain actions together except by applying a preset or style. You can, however, set up a single shortcut that does one thing in the lighttable view, say, and another in the darkroom view.
+
+**Additional modifiers.** As mentioned above, the only valid modifiers are the `Shift`, `Ctrl` and `Alt` keys on the keyboard. You can define additional keys (or device buttons) as modifiers by assigning keys/buttons to the "global/modifier" action. However, these will merely function as extra `Ctrl`, `Alt` or `Shift` keys -- you cannot create "new" modifiers.
```
paperdigits: I find this paragraph confusing. So I can add new keys, but they'll function as ctrl/alt/shift, but I can't add new modifiers?

paperdigits:
@masterpiga any fixes or comments for my comments?

paperdigits (changes requested):

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,126 @@ weight: 120
 draft: false
 ---
 
-You can perform almost any action in darktable with a keyboard/mouse shortcut. You can also use various other input devices, including MIDI devices and game controllers -- see the [midi device support](../special-topics/midi-device-support.md) section for details. These are referred to as _external devices_ or just _devices_ in this guide.
+_Shortcuts_ are a way to control darktable using your keyboard, pointing device or external controller. They allow you to perform _actions_ without interacting directly with a UI element.
```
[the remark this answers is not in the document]
masterpiga: Done.

content/preferences-settings/shortcuts.md
masterpiga: Done.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,135 @@ weight: 120
...
+
+When employing an input device, you can directly assign a control knob or joystick to an action. However, this will significantly reduce the flexibility of such devices, as you will be able to use the knob/joystick only to control one action. Conversely, by prefixing the movement with a key or button press you can use the same knob/joystick to control multiple actions.
+
+
+## triggering multiple shortcuts at once
+
+If both `e + scroll` and `f + scroll` are mapped to a slider (or if you have fallbacks enabled), then scrolling the mouse wheel while holding down both `e` and `f` will move both sliders.
+Similarly, you could map `e double-press + scroll` and `f double-press + scroll` to two nodes in a graph (say, `yellow` and `green` in color equalizer). You can then move both sliders at once by:
+
+1. pressing and releasing `e` once;
```
dterrahe: This should be a quick operation (<double click interval).

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,135 @@ weight: 120
...
+A _shortcut_ is a combination of inputs that triggers one of the actions available in a given context.
+
+Darktable comes with many predefined shortcuts using keyboard or keyboard and mouse, but you can also use various other input devices, including MIDI devices and game controllers -- see the [midi device support](../special-topics/midi-device-support.md) section for details. These are referred to as _input devices_ in this guide.
 
 The recommended way to assign shortcuts to visual elements is the [visual shortcut mapping](#visual-shortcut-mapping) mode.
 
-A single action may have multiple shortcuts but a single shortcut can only be linked to one action in a given darktable view -- you can't chain actions together except by applying a preset or style. You can, however, set up a single shortcut that does one thing in the lighttable view, say, and another in the darkroom view.
+# types of actions and shortcuts
+
+There are two types of actions:
```
dterrahe: I understand it can be useful to introduce concepts for didactic reasons that don't exactly match reality. I myself find the concept of continuous vs discrete actions confusing (because slider adjustments are clearly not "continuous"; they are intentionally stepwise so that "toggling" up/down will hit the same cache entries so you can quickly decide which you like best between two fine-tune steps. Truly discrete adjustments would make the cache useless for this). The _real_ difference is between "directional" shortcuts (up/down scroll, move etc) and "discrete" shortcuts (a press or click, double or whatever). For keyboard+mouse use you still need to be _holding_ your last "discrete" shortcut (which _must_ include a key press) in order to be able to combine it with a "directional" one. Since you are holding it anyway, it is easy and natural to issue a series of "directional" shortcuts from it. Does that make them "continous"? With external devices you can issue "directional" shortcuts independently.

The only reason to make the distinction is that it is natural to assign "directional" shortcuts to different "default" effects than "discreet" shortcuts. So while a key press shortcut linked to a slider will show its edit pop-up, a scroll will move it up. Or down. But those are only defaults; you can explicitly override them any way you want (both assigning discrete shortcuts to an up or adown effect or assigning one of the directions to a "discrete" effect (like the popup). With the current limitation that the other direction has to be assigned to the same action (may be different element/effect) but that's mostly to guarantee that they're both available on the same screen. Could be more generalised but I doubt there's demand.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,135 @@ weight: 120
 draft: false
 ---
 
-You can perform almost any action in darktable with a keyboard/mouse shortcut. You can also use various other input devices, including MIDI devices and game controllers -- see the [midi device support](../special-topics/midi-device-support.md) section for details. These are referred to as _external devices_ or just _devices_ in this guide.
+_Shortcuts_ are a way to control darktable using your keyboard, mouse or trackpad and other input devices. They allow you to perform _actions_ without interacting directly with a UI element.
```
dterrahe: I think I've said this before as well. I am not at all a gifted technical or manual writer so I don't want to second guess anybody who is. And the balance to be walked here is a fine one, I understand. Too many examples makes a messy text. This reference manual entry starts quite formal by introducing all the complicated concepts. It may make it less obvious to the more casual user that simple things are actually very simple. So they may conclude that this chapter is not what they are looking for. And maybe it isn't; maybe they should look elsewhere to see that if they simply want to link a key press to a screen button press they can do that in three/four steps:
1. press shortcut mapping button
2. hover over the button
3. press the key
(4. right click to switch off mapping mode)
Then once their use cases evolve from there, they might want to come back here to see what more is possible. I'm just worried all the complicated text here makes them believe that dt does _not_ have shortcut system that works for their needs at all.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,135 @@ weight: 120
...
+1. pressing and releasing `e` once;
+2. pressing and holding `e`;
+3. pressing and holding `f`;
+4. using the scroll wheel while both `e` and `f` are pressed.
+
+This works because the repeat key counter is per-shortcut, not per key, and it is reset only after the current shortcut ends (i.e., when the repeated key is released). Thus it is possible to trigger as many actions at once as your fingers can handle, provided that all the shortcuts involved have the same number of key presses (i.e., single, double or triple presses).
+
+## shortcuts must be unique within a view
+
+A single action may have multiple shortcuts but a single shortcut can only be linked to one action in a given darktable view -- you can't chain actions together except by applying a preset or style. You can, however, set up a single shortcut that does one thing in the lighttable view and another in the darkroom view.
```
dterrahe: ...except by applying a preset or style _or lua script_.

content/preferences-settings/shortcuts.md
```diff
@@ -103,23 +151,34 @@ The mouse cursor will change as you hover over UI widgets, to indicate whether o
 
 -   ![Don't signal](./shortcuts/no-signal.png#icon) indicates that there is no mappable widget under the cursor.
 
-Press a key combination while hovering over a mappable widget to assign a shortcut to that widget -- a default action will be assigned to that shortcut based on the type of widget and whether you have keyed a _simple_ or _extended_ shortcut. See below for details of some of the default assigned actions.
-
-Left-click on a mappable widget to open the shortcut mapping screen for that widget (see below). Left-click anywhere else on the screen to open the shortcut mapping screen, expanded (where possible) based on the part of the screen you have clicked on. This screen can be used to alter the action assigned to a shortcut and to configure shortcuts for non-visual actions. Entering the shortcut mapping screen exits visual shortcut mapping mode.
-
+**To define new shortcuts:**
+press a key combination while hovering over a mappable widget. A _default action_ will be assigned to that shortcut based on the type of widget and whether your shortcut includes a movement. See the [common actions](#common-actions) section for examples of some of the defaults.
```
dterrahe: "whether your shortcut includes a movement" you mean whether it is a continuous shortcut?

dterrahe (approved):
I added a few more commentaries but I didn't really thoroughly read through everything; there may be inaccuracies I didn't catch.

Since I'm far from the target audience I can't judge if this is an improvement overall or not. My "approval" here simply means I have no objection and don't want to prevent the manual maintainers from deciding.

content/preferences-settings/shortcuts.md
masterpiga: I would allow users to define custom actions implemented as ordered sequences of actions. If the action is directional, it can contain any number of directional actions and the same movement is applied to all of them. This would cover probably 90% of the use cases.

Also, I would not implement a UI for that. I would keep it as a "power user" feature and only implement the logic to parse these beasts from shortcutsrc.

content/preferences-settings/shortcuts.md
masterpiga: Done

content/preferences-settings/shortcuts.md
```diff
@@ -142,29 +201,30 @@ Alternatively, you may disable [preferences > miscellaneous > interface > load d
...
+For example, the predefined shortcut `tab` triggers the action `globals/panels/all`, which toggles the visibility of all the side panels in the current view.
 
-A button is a clickable icon in the darktable interface. The default action, when assigning a simple shortcut to a _button_, is to activate that button as if clicked with the left mouse button. You can modify this action to activate the button as if clicked while holding a modifier key.
+## view-specific actions
 
-## toggles
+Actions in the "views" section can only be executed from the specified darktable view. As with global actions, most do not have specific _elements_ as they are used to perform one-off operations.
 
-A toggle is a button that has a persistent on/off state. It therefore has additional _effects_ to allow you to toggle it or explicitly set its state. As with a normal button the default action, when assigning a simple shortcut to a toggle, is to activate the toggle as if clicked with the left mouse button (which toggles the button on/off).
+For example, the predefined shortcut `ctrl + b` triggers the action `views/darktable/guide lines/toggle`, which toggles guide lines in the darktable view.
```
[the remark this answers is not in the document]
masterpiga: Added

masterpiga:
Thanks, @dterrahe.

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
 draft: false
 ---
 
-You can perform almost any action in darktable with a keyboard/mouse shortcut. You can also use various other input devices, including MIDI devices and game controllers -- see the [midi device support](../special-topics/midi-device-support.md) section for details. These are referred to as _external devices_ or just _devices_ in this guide.
+_Shortcuts_ are a way to control darktable using your keyboard, mouse or trackpad and other input devices. They allow you to perform _actions_ without interacting directly with a UI element.
```
paperdigits:
```suggestion
_Shortcuts_ are a way to control darktable using your keyboard, mouse, trackpad, or other input devices. They allow you to perform _actions_ without interacting directly with a UI element.
```

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
-A _shortcut_ is a combination of key or button presses and/or mouse or device movements that performs an _action_ in darktable.
+-   Increase, decrease or reset sliders
+-   Scroll through dropdown lists
+-   Enable, expand or focus modules
+-   Click buttons
+-   Switch between views
+
+A _shortcut_ is a combination of inputs that triggers one of the actions available in a given context.
+
+Darktable comes with many predefined shortcuts using keyboard or keyboard and mouse, but you can also use various other input devices, including MIDI devices and game controllers -- see the [midi device support](../special-topics/midi-device-support.md) section for details. These are referred to as _input devices_ in this guide.
```
paperdigits: Please mention midi device support more near the first paragraph. This paragraph seems like a longer version of the first paragraphs. Can we merge them?

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
+1. Click on the ![visual mapping button](./shortcuts/visual-mapping-button.png#icon) icon
+2. Hover the slider that you want to control
+3. Press the key that you want to use to control the slider, and scroll the mouse wheel or move the mouse horizontally, vertically or diagonally
 
-## simple shortcuts
+The system is very flexible and powerful, and this is just scratching the surface.
+Keep on reading to understand it better and learn how to make the most of it.
 
-A shortcut that only includes button and/or key presses (and not mouse/device movements) is referred to as a _simple_ shortcut.
+# types of actions and shortcuts
```
paperdigits:
```suggestion
# types of shortcuts and actions
```

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
+- The key `e` can be used to focus a module, or to toggle it on/off
+- The key `e`, combined with up/down movements or the mouse, can be used to control the value of a slider
 
--   Movement of the mouse scroll wheel
--   Horizontal, vertical or diagonal movement of the mouse cursor
--   Movement of a knob/joystick on an external device
+---
+
+**Note:**
+By construction, a _directional shortcut_ is sufficiently expressive to trigger a _discrete action_. The opposite is not true. However, it is possible to use a combination of two _discrete shortcuts_ to control a _directional action_. For example, you can use two different shortcuts to trigger the `up` and `down` effects of a slider (see [anatomy of an action](#anatomy-of-an-action)).
```
paperdigits:
```suggestion
A _directional shortcut_ is sufficient to trigger a _discrete action_. The opposite is not true. However, it is possible to use a combination of two _discrete shortcuts_ to control a _directional action_. For example, you can use two different shortcuts to trigger the `up` and `down` effects of a slider (see [anatomy of an action](#anatomy-of-an-action)).
```

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
+* Three presses of the three mouse buttons, followed by
+* 2 more repetitions of the last click.
+
+Your shortcut can include one or more modifiers (`shift`, `ctrl` and `alt`). In this case, for a directional shortcut, both the last key and the modifier must be pressed down when you execute the movement. For example, `ctrl + e double-press + pan` means that while moving the mouse left and/or right both  `e` and `ctrl` must be pressed down.
 
 ---
 
-**Note:** You may need to switch off the "disable touchpad while typing" setting if you want to use extended shortcuts with a laptop touchpad.
+**Note:**
+If you are a MacOs user, your shortcuts will use `cmd` instead of `ctrl` and `option` instead of `alt`.
```
paperdigits:
```suggestion
If you are a macOS user, your shortcuts will use `cmd` instead of `ctrl` and `option` instead of `alt`.
```

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
+
+## movements in directional shortcuts
+
+The following movements are supported when defining directional shortcuts:
+
+-   Movement of the mouse scroll wheel
+-   Horizontal, vertical or diagonal movement of the mouse cursor
+-   Movement of a knob/joystick on an input device
+
+When employing an input device, you can directly assign a control knob or joystick to an action. However, this will significantly reduce the flexibility of such devices, as you will be able to use the knob/joystick only to control one action. Conversely, by prefixing the movement with a key or button press you can use the same knob/joystick to control multiple actions.
```
paperdigits:
```suggestion
When using an input device, you can directly assign a control knob or joystick to an action. However, this will significantly reduce the flexibility of such devices, as you will be able to use the knob/joystick only to control one action. By prefixing the movement with a key or button press you can use the same knob/joystick to control multiple actions.
```

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
 
 element
 : An _element_ is the part of a UI widget that is affected by your shortcut. For example, for a slider that has a [picker](../darkroom/processing-modules/module-controls.md#pickers), you can make a shortcut activate the picker _button_ element or change the _value_ element of the slider. For a row of tabs (the row is a single widget) you can select which tab element to activate or use your mouse scroll wheel to scroll through the tabs.
 
 effect
 : A shortcut can sometimes have multiple possible _effects_ on a given _element_. For example, a button can be activated as if it was pressed with a plain mouse-click or as if it was pressed with Ctrl+click. A slider's value can be edited, increased/decreased or reset.
 
+So, for example, you can assign the shortcut `e + scroll` to change the value (the _effect_) of the exposure slider (the _element_) of the exposure correction _widget_.
```
paperdigits:
```suggestion
For example, you can assign the shortcut `e + scroll` to change the value (the _effect_) of the exposure slider (the _element_) of the exposure correction _widget_.
```

content/preferences-settings/shortcuts.md
```diff
@@ -5,87 +5,154 @@ weight: 120
...
+
 # assigning shortcuts to actions
 
-There are two primary methods of assigning a shortcut to an action.
+There are two primary methods of assigning shortcuts to actions: [visual shortcut mapping](#visual-shortcut-mapping) (recommended) and the [shortcut mapping screen](#shortcut-mapping-screen). The former makes it easier to remap multiple shortcuts at once, while the latter allows for finer control and more flexibility. Both options are detailed below.
```
paperdigits:
```suggestion
There are two primary methods of assigning shortcuts to actions: [visual shortcut mapping](#visual-shortcut-mapping) (recommended) and the [shortcut mapping screen](#shortcut-mapping-screen). The former makes it easier to remap multiple shortcuts at once, while the latter allows for finer control and more flexibility.
```

content/preferences-settings/shortcuts.md
```diff
@@ -103,23 +170,34 @@ The mouse cursor will change as you hover over UI widgets, to indicate whether o
 
 -   ![Don't signal](./shortcuts/no-signal.png#icon) indicates that there is no mappable widget under the cursor.
 
-Press a key combination while hovering over a mappable widget to assign a shortcut to that widget -- a default action will be assigned to that shortcut based on the type of widget and whether you have keyed a _simple_ or _extended_ shortcut. See below for details of some of the default assigned actions.
-
-Left-click on a mappable widget to open the shortcut mapping screen for that widget (see below). Left-click anywhere else on the screen to open the shortcut mapping screen, expanded (where possible) based on the part of the screen you have clicked on. This screen can be used to alter the action assigned to a shortcut and to configure shortcuts for non-visual actions. Entering the shortcut mapping screen exits visual shortcut mapping mode.
-
+**To define new shortcuts:**
+press a key combination while hovering over a mappable widget. A _default action_ will be assigned to that shortcut based on the type of widget and whether it is directional. See the [common actions](#common-actions) section for examples of some of the defaults.
```
paperdigits:
```suggestion
Press a key combination while hovering over a mappable widget. A _default action_ will be assigned to that shortcut based on the type of widget and whether it is directional. See the [common actions](#common-actions) section for examples of some of the defaults.
```

content/preferences-settings/shortcuts.md
```diff
@@ -103,23 +170,34 @@ The mouse cursor will change as you hover over UI widgets, to indicate whether o
...
 You can assign as many shortcuts as you like in a single mapping session and then exit mapping mode when you are finished by clicking the ![visual mapping button](./shortcuts/visual-mapping-button.png#icon) icon again or right-clicking anywhere on the screen.
 
-You can delete a user-defined shortcut mapping by defining it a second time against the same widget. If you attempt to reallocate an existing shortcut to a new action, you will be notified of the conflict and asked whether you wish to replace the existing shortcut.
+**To explore already defined shortcuts:**
+left-click on a mappable widget to open the shortcut mapping screen for that widget (see below). Left-click anywhere else on the screen to open the shortcut mapping screen, expanded (where possible) based on the part of the screen you have clicked on. This screen can be used to alter the action assigned to a shortcut and to configure shortcuts for non-visual actions. Entering the shortcut mapping screen exits visual shortcut mapping mode.
```
paperdigits:
```suggestion
Left-click on a mappable widget to open the shortcut mapping screen for that widget (see below). Left-click anywhere else on the screen to open the shortcut mapping screen, expanded (where possible) based on the part of the screen you have clicked on. This screen can be used to alter the action assigned to a shortcut and to configure shortcuts for non-visual actions. Entering the shortcut mapping screen exits visual shortcut mapping mode.
```

content/preferences-settings/shortcuts.md
```diff
@@ -103,23 +170,34 @@ The mouse cursor will change as you hover over UI widgets, to indicate whether o
...
+
+**To update existing shortcuts:**
+define a second shortcut against the same widget. If you attempt to reallocate an existing shortcut to a new action, you will be notified of the conflict and prompted to replace the existing shortcut.
```
paperdigits:
```suggestion
Define a second shortcut against the same widget. If you attempt to reallocate an existing shortcut to a new action, you will be notified of the conflict and prompted to replace the existing shortcut.
```

content/preferences-settings/shortcuts.md
```diff
@@ -103,23 +170,34 @@ The mouse cursor will change as you hover over UI widgets, to indicate whether o
...
 
-Finally, if you scroll with your mouse wheel while in visual mapping mode (without pressing any other buttons/keys) when hovering over a slider, this will change the default speed for that slider -- scroll up to increase and down to decrease. When you leave mapping mode, normal mouse scrolls over that slider will change its value with the adjusted speed.
+**To change the default speed of a slider:**
+scroll with your mouse wheel while in visual mapping mode (without pressing any other buttons/keys) when hovering over the slider -- scroll up to increase and down to decrease. When you leave mapping mode, normal mouse scrolls over that slider will change its value with the adjusted speed.
```
paperdigits:
```suggestion
Scroll with your mouse wheel while in visual mapping mode (without pressing any other buttons/keys) when hovering over the slider -- scroll up to increase and down to decrease. When you leave mapping mode, normal mouse scrolls over that slider will change its value with the adjusted speed.
```

content/preferences-settings/shortcuts.md
```diff
@@ -103,23 +170,34 @@ The mouse cursor will change as you hover over UI widgets, to indicate whether o
...
+The top panel of the shortcut mapping screen shows a list of available UI widgets/actions and the bottom panel shows the shortcuts currently assigned to them. Fields that can be changed by user action are shown in bold.
 
-Double-click an item in the top panel to create a new shortcut for that item, and then enter your desired shortcut (right-click to cancel). Once you have done this, a new entry will appear in the bottom panel showing the shortcut you have created. You can then manually alter the _element_, _effect_, _speed_ or _instance_ of the assigned action against that shortcut in the bottom panel. To delete a user-defined shortcut, select it in the bottom panel and press the Delete key.
+---
+
+**Note:**
+You can **search** the top and bottom panels using the text entry boxes at the bottom of the screen, and use the **up/down** arrow keys to navigate between matches.
```
paperdigits:
```suggestion
You can search the top and bottom panels using the text entry boxes at the bottom of the screen, and use the up/down arrow keys to navigate between matches. 
```

content/preferences-settings/shortcuts.md
```diff
@@ -142,29 +220,30 @@ Alternatively, you may disable [preferences > miscellaneous > interface > load d
...
 
-The following is a list of some of the actions to which you can assign shortcuts, organized by widget type. This is not an exhaustive list and you are encouraged to browse the shortcut mapping screen for a complete list of available actions. If you assign a shortcut to a widget, it will be given a default action, depending on the type of widget and on whether you have assigned a simple or extended shortcut.
+The following is a list of some of the actions to which you can assign shortcuts, organized by widget type. This is not an exhaustive list and you are encouraged to browse the shortcut mapping screen for a complete list of available actions.
 
-Note that it is possible to assign a number of actions that have no effect. For example, all sliders include a _button_ element, regardless of whether such a button is actually present alongside a given slider.
+---
 
-## global
+**Note:**
+some actions may have no effect. For example, all sliders include a _button_ element, regardless of whether such a button is actually present alongside a given slider.
```
paperdigits:
```suggestion
Some actions may have no effect. For example, all sliders include a _button_ element, regardless of whether such a button is actually present alongside a given slider.
```

content/preferences-settings/shortcuts.md
```diff
@@ -173,47 +252,57 @@ _reset_
...
-
-Processing modules have the same elements and defaults as utility modules with the following additional elements:
+Additionally, processing modules have the following elements:
 
 _enable_
 : Acts as a _toggle_ that switches the module on and off.
 
 _focus_
-: Acts as a _toggle_ that focuses or defocuses the module. This is useful for modules such as [_crop_](../module-reference/processing-modules/crop.md) or [_tone equalizer_](../module-reference/processing-modules/tone-equalizer.md), whose on-screen controls are only activated when those modules have focus. For _crop_, changes are saved only when the module loses focus.
+: Acts as a _toggle_ that focuses or de-focuses the module. This is useful for modules such as [_crop_](../module-reference/processing-modules/crop.md) or [_tone equalizer_](../module-reference/processing-modules/tone-equalizer.md), whose on-screen controls are only activated when those modules have focus. For _crop_, changes are saved only when the module loses focus.
```
paperdigits:
```suggestion
: Acts as a _toggle_ that focuses or unfocuses the module. This is useful for modules such as [_crop_](../module-reference/processing-modules/crop.md) or [_tone equalizer_](../module-reference/processing-modules/tone-equalizer.md), whose on-screen controls are only activated when those modules have focus. For _crop_, changes are saved only when the module loses focus.
```

content/preferences-settings/shortcuts.md
```diff
@@ -226,26 +315,34 @@ _button_
...
 
-# fallbacks
+# fallbacks (a.k.a. generalized shortcut extensions)
 
-Where a _widget_ can have multiple different _actions_ applied to it, it can be tedious to set up individual shortcuts for each one of those actions. To make this process simpler, if you create a simple shortcut a number of effects can be made available by default as extensions to that shortcut. These are known as _fallbacks_.
+As discussed above, a _widget_ can have multiple different _actions_ applied to it, and you may want to set shortcuts to control several of them. For example, to toggle a module on/off, expand/collapse it, and reset it to its default state. Normally, this would require setting 3 shortcuts for each module. If you want to control 5 modules in this fashion, you have to define 15 shortcuts. Setting up all these shortcuts one by one can be tedious and error prone.
```
paperdigits, line 316: A _widget_ can have multiple different _actions_ applied to it, and you may want to set shortcuts to control several of them. For example, to toggle a module on/off, expand/collapse it, and reset it to its default state. Normally, this would require setting 3 shortcuts for each module. If you want to control 5 modules in this fashion, you have to define 15 shortcuts. Setting up all these shortcuts one by one can be tedious and error prone.
```

content/preferences-settings/shortcuts.md
masterpiga: Done.

content/preferences-settings/shortcuts.md
masterpiga: The comment is just a repetition of the suggested edit, isn't it?

masterpiga:
Thanks for all the suggestions, @paperdigits

paperdigits:
Thanks!

masterpiga:
Thank you all for your patience and careful review. It's an honor to contribute to dtdocs :)

Issue #794, opened 2025-12-05
Midi device support: Example wrong/misleading


markusb:
Hello Darktable,

in the page documenting midi device support (special-topics/midi-device-support) there is an example about configuring the field "preferences > miscellaneous > interface > order or exclude midi devices". The example is using semicolons to separate devices "`BeatStep;dontuse;X-TOUCH;-BCR2000`". Unfortunately this is wrong, the correct way to separate devices is a simple colon. The sample should read "`BeatStep,dontuse,X-TOUCH,-BCR2000`".

I have tested this in darktable 5.2.1, with a semicolon it does not work, but with a simple colon it works.

Can you please fix this?

Thanks you

Markus

PR #795, opened 2025-12-07
Fix example of midi device support config to use commas.

Phemisters:
Related to dtdocs issue: https://github.com/darktable-org/dtdocs/issues/794

The  `preferences > miscellaneous > interface > order or exclude midi devices` config string uses commas to separate devices, not semi-colons (and not colons, they are used within a field for encoding config).


elstoc:
@dterrahe please can you review this

PR #799, opened 2025-12-19
AgX docs

kofa73:
https://github.com/darktable-org/darktable/pull/19026
(and a few more)

New page in the manual


victoryforce:
We are having some issues displaying [this page](https://docs.darktable.org/usermanual/development/en/module-reference/processing-modules/agx/) by Hugo.

As you can see, site theme adds the following text to the H4 heading on this page: "We only allow H1, H2, or H3 headings in body content".

What do you think would be better: removing this warning from the site's theme or perhaps somehow restructuring the documentation page to avoid such a depth of headings?

PR #810, opened 2026-01-14
Document unsupported file formats in darktable

ghost:
Added a section listing explicitly unsupported file formats.

### DIFF
--- a/content/overview/supported-file-formats.md
+++ b/content/overview/supported-file-formats.md
@@ -25,3 +25,16 @@ With the exception of Fujifilm X-Trans cameras, darktable does not decode images
 darktable natively reads “ordinary” images in JPEG, 8-bit/16-bit PNG and 8-bit/16-bit TIFF format, as well as 16-bit/32-bit floating point TIFF formats.
 
 darktable also reads high dynamic range images in OpenEXR, RGBE and PFM formats.
+
+
+# explicitly not supported
+
+The following formats are explicitly not supported:
+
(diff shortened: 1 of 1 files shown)


elstoc:
I think I would prefer to just link to the camera support page (which will always be more up-to-date than here). Honestly the list of extensions in this section isn't really the most useful thing anyway. Maybe just append "...at time of writing (please see **link** for an up-to-date list of supported cameras)" to the first sentence

ghost:
The list of supported camera's is auto generated. But that list of explicit unsupported is not. It would be strange to link outside the document for this, as the documentation is the source of truth. I would actually make sense that article would [Camera Support](https://www.darktable.org/resources/camera-support) would link to the documentation

paperdigits:
The original source of truth for this information is the release news post on the website.

elstoc:
The user manual isn't the only source of "documentation about darktable". It is primarily concerned with "how darktable works" rather than specifics of support for operating systems or cameras (for example). I'd like to avoid the user manual needing to be updated frequently, which we would have to do with detailed camera/file support info.

As I say I don't mind putting this particular set of information in, as long as we make it clear it might not be up-to-date, and link to where the up-to-date information is.

paperdigits:
I'd much rather say something like "you can find a list of specifically unsupported formats on the camera support page." or something like that, and link to the news category. I already manually copy the unsupported file types from the news to the camera support page, I don't want to copy it to a third place.

PR #811, opened 2026-01-14
Can use shift modifier to select feathering node in masks

Phemisters:
This relates to this darktable PR: https://github.com/darktable-org/darktable/pull/18896

User can now use Shift when selecting a feather outline node to avoid selecting an adjacent mask node.

The same PR altered the modifier for changing an ellipse mask's feather mode (from Shift to Alt), so update that in doc as well. (Note that the help text at the top of screen when an ellipse mask is selected has not been updated - I've opened a PR about that: https://github.com/darktable-org/darktable/pull/20155).


Phemisters:
> I didn't realize @Phemisters already did this documentation-pending item (first time here), and so independently checked this out - I agree with the changes. Also just saw you already fixed the incorrect documentation string already ;-)

Sorry to overlap with your changes - but it is great seeing quite a few people contributing to the docs! Hope you find another issue to work on.

PR #819, opened 2026-01-19
update: tidy up all front-matter

### LINKED ISSUE #816: Tidy up front-matter metadata
We're using front-matter metadata in a rather inconsistent way and often not keeping it up-to-date.

IMO we should just stick to the front-matter that's required in order to make the pages functional, and strip out all except **title** **id** and **weight**

Would be great if we could find a way to automate this...

ghost:

Closes #816

Tidies up all front-matter, keeping only:
- **title**
- **id**
- **weight**
- **include_toc**


elstoc:
Nice one. Many thanks!

PR #823, opened 2026-01-24
Add doc on editing global color picker live samples

Phemisters:
Relates to darktable PR: https://github.com/darktable-org/darktable/pull/17051

Adds a description of how to edit a live sample in the global color picker.

I found it a little difficult at first to understand the change, but two of the PR comments helped a lot: [here](https://github.com/darktable-org/darktable/pull/17051#issuecomment-2474375934) and [here](https://github.com/darktable-org/darktable/pull/17051#issuecomment-2479413765). To assist with the review, this is extracted from the first of those comments:

> * Hover over a live sample
> * Right-click on the sample
> * The live sample is now copied to the primary sample
> * Change the size/positon of the sample
> * Hover over a live sample (not necessary the same used above)
> * Right-click on the sample
> * The live sample is receiving the primary sample


elstoc (commented):
This being a bulleted list, I can't work out whether it's three different options, or a list of actions that needs to be executed in order. If the latter, it should probably be a numbered list. Or I guess it could just be a single multi-sentence paragraph?

Phemisters:
> Or I guess it could just be a single multi-sentence paragraph?

@elstoc Thanks for the suggestions.  I have tried editing it to be a single paragraph. Do you think that is better?

elstoc:
Yeah, much clearer, thanks.

PR #825, opened 2026-01-25
Add import extras description to shortcuts doc

Phemisters:
Relates to darktable PR: https://github.com/darktable-org/darktable/pull/19500

Add a description of the `import extras` button in the shortcuts preferences page.

This doc change is adapted from another dtdocs PR, which was closed because it  added other, unwanted, details:  https://github.com/darktable-org/dtdocs/pull/783

Note that github reports an additional change on the last line of the file, but that is due to my changed file having a final line-feed at the end. I think is normal enough for the final line to be terminated by a line-feed (other dtdocs files generally end with a line-feed), and I have not tried to fix that.


elstoc:
Thanks!

PR #830, opened 2026-02-09
Update crop doc with aspect ratio display

Phemisters:
Relates to this darktable PR: https://github.com/darktable-org/darktable/pull/20191

The darktable change displays the crop area aspect ratio during the crop area resizing. 

I don't think we currently mention in the doc that the pixel size is already shown, so I've added a sentence mentioning both. It might be that we decide mot to bother mentioning what is displayed in the centre, since it is probably obvious, in which case we can just close this doc PR.


elstoc:
LGTM Thanks

PR #831, opened 2026-02-09
Can switch picker mode with ctrl+click

Phemisters:
Relates to this darktable PR: https://github.com/darktable-org/darktable/pull/19827

It is now possible to switch the picker mode (point or area) by a Ctrl+click on the image.


elstoc:
Nice one, thanks

PR #833, opened 2026-02-11
Add vectorize button to external raster masks doc

Phemisters:
Relates to this darktable PR: https://github.com/darktable-org/darktable/pull/20198

Add the new _vectorize_ button to the _external raster masks_ module documentation. This allows users to generate path masks from raster mask files.

For experimenting, I used the PFM file linked to in [one of the PR comments](https://github.com/darktable-org/darktable/pull/20198#issuecomment-3792846239).

I also noticed that the text for a preference related to external raster masks was incorrect, so I updated it to match the current preferences text.

### DIFF
--- a/content/module-reference/processing-modules/external-raster.md
+++ b/content/module-reference/processing-modules/external-raster.md
@@ -21,7 +21,10 @@ Using this module, external raster mask files are made available as a raster mas
 # module controls
 
 file selection
-: Choose the external raster file to use. File selection is inactive if the rasterfile root folder has not been defined in [preferences > processing](../../preferences-settings/processing.md).
+: Choose the external raster file to use. File selection is inactive if the raster mask files root folder has not been defined in [preferences > processing](../../preferences-settings/processing.md).
 
 mode
 : Choose any of the RGB channels, or a combination of them, to generate the raster mask from.
+
+vectorize
+: Create a separate path mask for each shape identified in the current raster mask, and add all the resulting path masks to the mask manager.
--- a/content/module-reference/processing-modules/external-raster.md
+++ b/content/module-reference/processing-modules/external-raster.md
@@ -21,7 +21,7 @@ Using this module, external raster mask files are made available as a raster mas
 # module controls
 
 file selection
-: Choose the external raster file to use. File selection is inactive if the raster mask files root folder has not been defined in [preferences > processing](../../preferences-settings/processing.md).
+: Choose the external raster file to use. The initial raster mask files root folder can be set in [preferences > processing](../../preferences-settings/processing.md).
 
 mode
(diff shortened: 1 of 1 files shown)


content/module-reference/processing-modules/external-raster.md
```diff
@@ -21,7 +21,10 @@ Using this module, external raster mask files are made available as a raster mas
 # module controls
 
 file selection
-: Choose the external raster file to use. File selection is inactive if the rasterfile root folder has not been defined in [preferences > processing](../../preferences-settings/processing.md).
+: Choose the external raster file to use. File selection is inactive if the raster mask files root folder has not been defined in [preferences > processing](../../preferences-settings/processing.md).
```
anoderay: I would suggest replacing "inactive" with "might be unavailable". I am also not even sure it is true. I set the raster mask folder to blank in darktablerc to try this out and could still load a raster mask without issue.

Phemisters:
@anoderay I'm not sure why you assigned this to me.

TurboGit:
@anoderay : Maybe for a final quick review (check that only needed files are in the PR) and merged.

I have asked for the final merge to be done by another contributor if possible (except for quick typos). But that's only my proposal, you can of course agree on a workflow for all contributors to the doc.

anoderay:
> @anoderay I'm not sure why you assigned this to me.

You are correct, that didn't make sense. 

Did you see my comment?

Phemisters:
> Did you see my comment?

I don't see any comments. If you started a review, you need to submit  the review before I see the comments.

anoderay (changes requested):
Here's the comment

content/module-reference/processing-modules/external-raster.md
Phemisters: I've pushed an update. Do you think that is better?

victoryforce: > I would suggest replacing "inactive" with "might be unavailable"

Please don't do this! I suggest adding the word "might" to the blacklist of words that are prohibited in the user guide. "Might" is a word used as a modal verb to express possibility or uncertainty, often suggesting a lower likelihood than "may". 

> I am also not even sure it is true. I set the raster mask folder to blank in darktablerc to try this out and could still load a raster mask without issue.

Yes, you're right. I assume that this description was copied from the description of a similar control element of the LUT3D module. But since then, changes have been made and the LUT3D root directory in the preferences now has a default value (i.e. it simply cannot be undefined), as a result the description of "file selection" in the LUT3D module became inaccurate (it also needs to be corrected). The external raster masks module has a similar implementation, its root folder always has some definition and the control cannot be inactive.

anoderay: Thank you for this important note @victoryforce . @phemisters I think the wording from your latest update is fine.

PR #838, opened 2026-02-24
Adds information about pinned images in the second darkroom window

masterpiga:
Following the merge of https://github.com/darktable-org/darktable/pull/19963

### DIFF
--- a/content/darkroom/darkroom-view-layout.md
+++ b/content/darkroom/darkroom-view-layout.md
@@ -64,7 +64,7 @@ From left to right:
 : Quick access menu for styles. Hover over a style name with your mouse to show a preview of the current darkroom image with the selected style applied.
 
 ![second-window-icon](./darkroom-view-layout/second-window-icon.png#icon) second darkroom window
-: Allows the user to display a second image in a separate window. By default, the second window mirrors the contents of the main view. However, any point in the history of the image currently being developed can be 'pinned', causing the image in the second window to persist when selecting a different history point or a different image. This is especially useful on large screens and multi-monitor setups.
+: Allows the user to display a second image in a separate window. By default, the second window shows the same contents as the main view. When an image is "pinned" in the  second window, it will remain unchanged when making module changes, navigating the history stack or opening a different image in the main view. This is especially useful on large screens and multi-monitor setups.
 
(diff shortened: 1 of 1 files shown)


masterpiga:
@elstoc @paperdigits who is the right person to ask to review this? Thanks!

content/darkroom/darkroom-view-layout.md
```diff
@@ -64,7 +64,7 @@ From left to right:
 : Quick access menu for styles. Hover over a style name with your mouse to show a preview of the current darkroom image with the selected style applied.
 
 ![second-window-icon](./darkroom-view-layout/second-window-icon.png#icon) second darkroom window
-: For multi-monitor setup, allows the user to display a second image on another screen.
+: Allows the user to display a second image in a separate window. By default, the second window mirrors the contents of the main view. However, any point in the history of the image currently being developed can be 'pinned', causing the image in the second window to persist when selecting a different history point or a different image. This is especially useful on large screens and multi-monitor setups.
```
Phemisters: I wonder if `the second window mirrors the contents of the main view.` would be better as `the second window mirrors the current darkroom image.` What do you think?

lefth: I think "mirror" doesn't fit "content" (to my ears, an object mirrors an equal object, not content). I would suggest we simplify it to `The second window shows the same image as the main view. When the second window's image is "pinned", it will remain unchanged when making module changes or opening a different image.`

Phemisters: I'm okay either way - we talk about disk mirroring for instance, which is about replicating content (and updating in a synchronised way).

Phemisters: @masterpiga I'm new to reviewing dtdocs PRs, so thought perhaps I should have mentioned you by name after leaving some comments :smile:

masterpiga: Thanks, @Phemisters! Sorry for the delay, this had flown under my radar for some reason.

I rephrased as follows:

Allows the user to display a second image in a separate window. By default, the second window shows the same contents as the main view. When an image is "pinned" in the  second window, it will remain unchanged when making module changes, navigating the history stack or opening a different image in the main view. This is especially useful on large screens and multi-monitor setups.

Phemisters (approved):
Looks okay to me.

masterpiga:
Thank you for your review, @Phemisters!

PR #840, opened 2026-03-03
add info about new preview combobox

deekayhd:
This is a proposal for the documentation of a new option in the styles module introduced with PR [#20255](https://github.com/darktable-org/darktable/pull/20255).
The PR introduces a new combo box in the styles module for the preview on the tool tip. There are three options: default, large, no. With "no" it is possible to hide the preview.


anoderay:
Looks fine to me. Thanks!

PR #842, opened 2026-03-11
Watermark: add note about effect of framing on position

### LINKED ISSUE #724: Expand documentation on using a watermark and framing
Default module order puts framing (1) before the watermark (2). As a consequence the watermark is referenced on the framing and not the image.

Image

In some instances (my use case) I would like to reference the watermark on the image independent of the framing I choose. Therefore, in my case, I would like to place watermark (1) before framing (2). 

Image

Maybe a pointer to the fact that this result is achievable by changing the module order would be beneficial in the documentation.

deekayhd:
This is a proposal for issue #724.

I added a note to the documentation of the watermark module that the framing module affects the positioning of the watermark and that this can be solved by changing the default order of the modules manually.

closes #724

### DIFF
--- a/content/module-reference/processing-modules/watermark.md
+++ b/content/module-reference/processing-modules/watermark.md
@@ -72,6 +72,6 @@ y offset
 
 ---
 
-**Note**: When using the watermark in combination with the framing module, the added frame affects the watermark’s position. For example, if you align the watermark to an edge, it may appear within the frame instead of on the actual image. To ensure the watermark is positioned relative to the image itself, you can manually manually [adjust the module order](../../../darkroom/pixelpipe/the-pixelpipe-and-module-order.md#changing-module-order) so that the framing module comes after the watermark module in the pixelpipe.
+**Note**: When using the watermark in combination with the framing module, the added frame affects the watermark’s position. For example, if you align the watermark to an edge, it may appear within the frame instead of on the actual image. To ensure the watermark is positioned relative to the image itself, you can manually manually [adjust the module order](../../darkroom/pixelpipe/the-pixelpipe-and-module-order.md#changing-module-order) so that the framing module comes after the watermark module in the pixelpipe.
 
 ---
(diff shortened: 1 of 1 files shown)


content/module-reference/processing-modules/watermark.md
```diff
@@ -69,3 +69,9 @@ x offset
 
 y offset
 : Pixel-independent offset relative to the choice of alignment on the y-axis.
+
+---
+
+**Note**: When using the watermark in combination with the framing module, keep in mind that the added frame affects the watermark’s position. For example, if you align the watermark to an edge, it may appear within the frame instead of on the actual image. To ensure the watermark is positioned relative to the image itself, you can manually adjust the module order so that the framing module comes after the watermark module in the pixelpipe.
```
Phemisters: This looks good. I have a couple of suggestions for you to consider:

1. maybe replace `keep in mind that the added frame` with `the added frame`, as I guess that will be easier for translations.
2. adjusting the module order is maybe not that common, so maybe link to it with something like
```
manually [adjust the module order](../../../darkroom/pixelpipe/the-pixelpipe-and-module-order.md) so
```

Phemisters: @deekayhd I'm new to reviewing dtdocs PRs, so thought perhaps I should have mentioned you by name after leaving some comments :smile:

deekayhd: @Phemisters Never mind, I got a notification, anyway :-) Your suggestions make sense to me, however, I will not be able to make the changes within the next two week. I will take care of it then.

deekayhd: @Phemisters Now I made the changes you suggested. I slightly adjusted the link for the module order to directly jump to the relevant section in that document.

Phemisters: Thank you.

Issue #776, opened 2025-08-20
The links on this page to the Darktable user manual...give error 404...


hlgraff123-blip:
Here is the page with the dead links to the user manual:  https://docs.darktable.org/usermanual/development/en/ 

Please fix the links.

Thanks

H

ronanmccoy:
Came here for the same issue. The links are still broken. An explanation, and possible workaround, was posted in https://github.com/darktable-org/dtdocs/issues/771.

syborgxx:
Hello. Been a few changes in 5.4 and I can't find the pdf manual to download.

paperdigits:
We are not currently producing a PDF for 5.4.

paperdigits:
This is the most up to date we have. AgX docs are still in a PR.

Phemisters:
PR https://github.com/darktable-org/dtdocs/pull/850 has been merged, and should enable the building of the PDF and ePub versions for the [development version of the manual](https://docs.darktable.org/usermanual/development/en/) (which is the only one being updated at present).

There are some warnings (and possibly some formatting issues) which will be dealt with separately.

PR #850, opened 2026-05-09
Try to enable PDF and ePub creation

Phemisters:
Relates to: #776

There have been a number of requests for PDF and ePub versions to be available again. I tried some experiments locally and these changes do seem to allow these versions to be created (possibly with some errors, which can be looked at separately).

@TurboGit - a few questions:

1. is this worth doing?
2. do you know who might be able to review this?
3. does the deployment of the generated html, PDF, and ePub versions happen automatically (a github action of some kind)?

### DIFF
--- a/.github/workflows/pr-tests.yml
+++ b/.github/workflows/pr-tests.yml
@@ -26,7 +26,7 @@ jobs:
         run: nix develop --command bash -c "cd tools && bash generate-translations.sh --no-update"
 
       - name: Build epub
-        run: nix-shell --run './tools/build-epub.sh'
+        run: nix develop --command bash -c "cd toold && bash build-epub.sh"
 
       - name: Clear
         run: nix develop --command bash -c "rm -rf public"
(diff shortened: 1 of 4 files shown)


Phemisters:
> 3\. does the deployment of the generated html, PDF, and ePub versions happen automatically (a github action of some kind)?

I see that this involves [gh-pages.yml](https://github.com/darktable-org/dtdocs/blob/master/.github/workflows/gh-pages.yml), so that would need updated if we do start generating the different formats.

Phemisters:
I did some experiments with the workflow which runs tests on PRs to see if PDF and ePub files were created, and they were. I altered the `pr-tests.yml` file to save artefacts; if we merge this we can turn off the artefact saving.

I *think* because `gh-pages.yml` uses the `build-all.sh` script and seems to deploy everything in `public` that it might just work without further changes, but hard for me to be sure.

anoderay:
Do we have an easy way back when we merge this? Should we just gun it?

Phemisters:
> Do we have an easy way back when we merge this? Should we just gun it?

I'm not sure exactly how the deployment works. I can, however,  see no reason why reverting this change would fail fix any problem it caused.  So, yes, I *think* we could just merge it and see :see_no_evil:

anoderay:
Yolo

The tests for building pdf and epub passed so I think we should be fine.

Phemisters:
> The tests for building pdf and epub passed so I think we should be fine.

It does look like it has deployed something. I think there may still be some formatting issues and warnings, but I'll deal with them separately. Thanks for merging!

anoderay:
Neat, the links on the official mirror https://docs.darktable.org/usermanual/development/en/ also work. Thanks for taking care of that @Phemisters !

PR #852, opened 2026-05-09
Added documentation for config folder location and contents

anoderay:
Looking through various parts of the docs I felt like it was missing a central explanation of the config-folder's location and its contents. I think this would also be useful to reference/link to in other parts of the docs.

Since I am a mere advanced user please especially review for technical accuracy.

### DIFF
--- a/content/preferences-settings/config-folder.md
+++ b/content/preferences-settings/config-folder.md
@@ -9,16 +9,17 @@ darktable stores its settings, presets, styles, library database, etc. in a defa
 * Linux 
     + Installed via package manager or AppImage: `$HOME/.config/darktable`
     + Flatpak: `$HOME/.var/app/org.darktable.Darktable/config/darktable`
+    + snap: `$HOME/snap/darktable/current/`
 * Windows: `%LOCALAPPDATA%\darktable`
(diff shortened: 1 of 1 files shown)


tgz-code:
I can confirm this is correct on Windows. @Phemisters, @tpinfold,  can you confirm for other OS?

tgz-code:
I will say that most of the information on where these config files are stored is in the manual, but not centralized at all. I think it would be helpful to have a dedicated place in the manual explaining where config data is stored.

Phemisters:
> I can confirm this is correct on Windows. @Phemisters, @tpinfold, can you confirm for other OS?

I can only confirm that the Linux package manager one looks correct - I don't use the other methods.

anoderay:
The locations are correct, I checked each variant (except macOS). Are you guys fine with the file/folder explanations?

Phemisters:
> Are you guys fine with the file/folder explanations?

We might want to expand some of them later, but I'm fine with how it is.

lefth:
I can confirm on Mac OS the path is correct.

tpinfold:
I can confirm this is correct for windows and feel it should be merged.

PR #856, opened 2026-05-10
add link to the comprehensive darktable lua docs

anoderay:
Should be self explanatory. Always found it confusing the darktables lua docs aren't linked in dtdocs? @wpferguson any objections?


wpferguson:
@anoderay my goal was/is to write a darktable Lua programmers guide and just add a reference to it in the darktable docs.  But at the rate I'm going (or not going) a link is a good idea :smile:

Issue #847, opened 2026-05-09
Documentation proposal for the new message log history


TurboGit:
Merged PR: https://github.com/darktable-org/darktable/pull/20956

> The log history viewer provides a persistent record of application log messages displayed in the bottom panel (e.g., export/import progress, library operations, warnings, and errors). To access it, click the speech-bubble icon in the center-bottom-right toolbar. A scrollable popover opens showing all stored messages with timestamps.
> Key features:
> 
> - Messages are stored in a circular buffer of 1000 entries; when the limit is reached, older entries are removed as new ones arrive
> - Consecutive identical messages are automatically deduplicated (e.g., repeated "exporting image" updates)
> - Entries are displayed in monospace format with HH:MM:SS timestamps
> - Individual entries are selectable for copy-paste
> - Messages are loaded dynamically when the popover opens
> - New messages are not added to the visible popover automatically (refresh by toggling the popover)

Phemisters:
@TurboGit I'm assuming your intention is that someone opens a PR for the suggested change? Before I do that, I do want to check a few things:

1. the context menu has _cut_, _paste_, _delete_ - should these be removed as they don't seem appropriate? Or do they do something that I don't notice?
2. the _select all_ context menu item highlights a single line, which is not what _select all_ would normally do - should the menu text maybe be changed?
3. when it says _Messages are loaded dynamically when the popover opens_, that sounds like an implementation detail. From the user's point-of-view it does not matter when it is populated- as the next line says, they have to toggle the popover to see new messages.

TurboGit:
The description above was indeed more a set of hints to write proper documentation.

The last point is wrong in latest implementation BTW.

1. Indeed, I didn't notice there was a context menu, this should be removed (I'll fix).
2. Should be removed, see point 1.
3. Fully agreed this is implementation details.

TurboGit:
@Phemisters : The contextual menu is now removed.

anoderay:
Closed by https://github.com/darktable-org/dtdocs/pull/857

PR #857, opened 2026-05-10
Doc for view log history

Phemisters:
Relates to #847 and darktable PR https://github.com/darktable-org/darktable/pull/20956.

Adds some doc for the new _view log history_ button on the bottom panel. Text is adapted from issue #847.

Before this change the bottom panel controls were (I think) all specific to a view, so I added a new page for this first common control. I'm open to suggestions of better ways to add it.


TurboGit (commented):
For the online help system from Darktable app we add:

`  dt_gui_add_help_link(d->button, "message_log");`

Meaning that when I ask help on the UI using the `?` icon I jump to the "message_log" section. I don't know how this is materialized on the doc, but somehow it is required to do something to ensure that the proper section is opened.

Phemisters:
> For the online help system from Darktable app we add:
> 
> ` dt_gui_add_help_link(d->button, "message_log");`
> 
> Meaning that when I ask help on the UI using the `?` icon I jump to the "message_log" section. I don't know how this is materialized on the doc, but somehow it is required to do something to ensure that the proper section is opened.

It looks like this is handled in darktable. The `src/common/usermanual_url.c` file has a list of mappings and we'd need something like:

```
  {"message_log",                "overview/user-interface/bottom-panel/#view-log-history"},
```

The right side of the mapping has to match whatever gets committed in the docs - I'm happy to change the page or naming to suit you.

It is possible to experiment with a local manual by altering the `darktablerc` file, like this

```
context_help/last_url=http://localhost:1313/usermanual/development/
context_help/url=http://localhost:1313/usermanual/
context_help/use_default_url=false
```

TurboGit:
> The right side of the mapping has to match whatever gets committed in the docs

Yes, exactly. This is anchor `#<name>` that needs proper encoding on the documentation. But since the doc is in markdown and the final documentation an HTML page my point was : I don't know what should be done on the markdown documentation to have the proper tag on the HTML side after conversion.

Phemisters:
> I don't know what should be done on the markdown documentation to have the proper tag on the HTML side after conversion.

Ah, now I understand. My previous suggestion was not quite right. The anchors are only generated for headings, so for instance we see an existing `usermanual_url.c` entry is:

```
  {"global_toolbox",             "overview/user-interface/top-panel/#on-the-right-hand-side"},
```

which goes to the **heading** _"on the right hand side"_ in the top panel page, not to a specific control description.

So, if we use the markdown I have in this PR:

```
# common controls
```

 the usermanual_url.c link would be

```
  {"message_log",                "overview/user-interface/bottom-panel/#common-controls"},
```

Since there is currently only one common control, if you prefer I can change the heading to be more specific.

content/overview/user-interface/bottom-panel.md
anoderay: It may be due to tiredness but I found this sentence confusing. Maybe
`The bottom panel appears in all views. Common controls are described here; view-specific controls (e.g. darkroom mode) are described in the appropriate pages.`

Also I think it would make sense to link to the view-specific descriptions if available.

anoderay (changes requested):

content/overview/user-interface/bottom-panel.md
Phemisters: @anoderay thanks - I've pushed some changes following your suggestions.

anoderay (approved):

Phemisters:
@TurboGit The dtdocs change has been merged. Who normally makes the corresponding change to `usermanual_url.c` - is it the `dtdocs` author person or the `darktable` developer? I think we need to add:

```c
 {"message_log",                "overview/user-interface/bottom-panel/#common-controls"},
```

TurboGit:
> @TurboGit The dtdocs change has been merged. Who normally makes the corresponding change to `usermanual_url.c` - is it the `dtdocs` author person or the `darktable` developer? I think we need to add:
> 
> ```c
>  {"message_log",                "overview/user-interface/bottom-panel/#common-controls"},
> ```

I'll do it.

TurboGit:
Done and it works fine. Thanks.

Issue #809, opened 2026-01-13
Enable Hugo's page ToC


paperdigits:
I thought I did this a while ago, but we should enable hugo's page ToC. Seems like it'd be useful. https://gohugo.io/methods/page/tableofcontents/

elstoc:
yes we decided to only include it on the really complex pages

elstoc:
As discussed, `include_toc` is used in "filmic rgb" and "color calibration" but doesn't render for some reason

Phemisters:
Hopefully fixed by https://github.com/darktable-org/dtdocs/pull/866

PR #866, opened 2026-05-13
Enable in-page table of contents to display

Phemisters:
Relates to dtdocs issue: https://github.com/darktable-org/dtdocs/issues/809

A dtdocs issue was raised a while ago about the in-page _table of contents_ not displaying. The _filmic rgb_ page, for instance, has table of contents enabled (`include_toc: true`), but nothing displays.

Looking at the generated html, I see that there _**is**_ a table of contents div created, but the anchor tag to expand it has no text, so nothing is displayed.

This PR adds a default translation which means that _Table of Contents_ is displayed, and clicking it expands or collapses the table.

It is not used in many places, but we may as well have it working.


anoderay:
I suppose you did local testing and found no issues? Happy to merge if yes, easy to revert if we find issues.

Phemisters:
> I suppose you did local testing and found no issues?

Yes I did do local testing, and didn't see any issues. I generally do local testing by running something like `hugo server --disableFastRender` to allow to me to see what the pages look like.

anoderay:
Alright then let's go.

Issue #868, opened 2026-05-13
fix (near) infinite loading


anoderay:
At least with Firefox and Chromium I get a near infinite loading indicator even though the page seems to have fully loaded. I think this is related to https://piwik.kees.nl being called in [analytics.html](https://github.com/darktable-org/dtdocs/blob/master/themes/hugo-darktable-docs-theme/layouts/partials/analytics.html). This seems to have been a piwik instance run by andabata. 

<img width="1002" height="363" alt="Image" src="https://github.com/user-attachments/assets/394627f0-44df-4855-aca1-6d9e71e4a16a" />

@Phemisters do you think it is safe to simply remove the file?

Phemisters:
> do you think it is safe to simply remove the file?

@anoderay good question. I'm guessing we don't have access to any analytics.  I'd be okay with trying to remove `analytics.html` (and the reference to it in `themes/hugo-darktable-docs-theme/layouts/_default/baseof.html`) and see if it causes any issues. Analytics can be looked at later, if we think it is needed.

anoderay:
Closed by https://github.com/darktable-org/dtdocs/pull/869

PR #869, opened 2026-05-13
fix infinite loading because of dead analytics

anoderay:
Analytics seemed to run on a piwik instance which is (presumably permanently) offline. See https://github.com/darktable-org/dtdocs/issues/868 . 

This pr removes the analytics.html and corresponding reference in the layout. 

Tested on local hugo instance without issues.


Phemisters (approved):
I tried it locally and did not see any issues.

PR #871, opened 2026-05-14
specified blur-radius range

anoderay:
Adresses https://github.com/darktable-org/darktable/pull/20993 by specifying the blur-radius-range.

### DIFF
--- a/content/module-reference/processing-modules/blurs.md
+++ b/content/module-reference/processing-modules/blurs.md
@@ -21,7 +21,7 @@ A diagram at the top of the module shows the shape of the blurring operator (kno
 ## general
 
 blur radius
-: The spreading size of the blur.
+: The spreading size of the blur (4-256 px).
 
 blur type
 : Choose between the different blur variants (above).


content/module-reference/processing-modules/blurs.md
```diff
@@ -21,7 +21,7 @@ A diagram at the top of the module shows the shape of the blurring operator (kno
 ## general
 
 blur radius
-: The spreading size of the blur.
+: The spreading size of the blur (4-256 px).
```
Phemisters, line 24: I don't know enough about this to know if the radius is pixels or someting else. Can you confirm that you know it is pixels?

victoryforce:
I'm not sure if we should even start specifying value ranges in the description of controls? If we do here, then why don't we do it consistently throughout the documentation?

anoderay:
@victoryforce I was hesitant to add the information as well. Someone manually entering 1000 is a cornercase anyway and then they would naturally find out the hard limit. I would be absolutely fine with closing this pr.

content/module-reference/processing-modules/blurs.md
anoderay: Yeah it is pixels, visible in the UI.

victoryforce:
Also, "`documentation: pending`" is not a mandatory requirement to write something in the documentation. This is more of a "note to remember" that the PR makes changes that _potentially_ should be reflected in the documentation (or maybe not, if the documentation maintainers think it shouldn't).

PR #873, opened 2026-05-15
added instructions per OS how to invoke darktable

anoderay:
The docs were missing OS-specific instructions on how to invoke darktable and the other binaries like darktable-cltest. 

Cannot test mac, would be grateful if someone else could do that. I did test snap, flatpak, appimage, package and windows.

### DIFF
--- a/content/special-topics/program-invocation/darktable.md
+++ b/content/special-topics/program-invocation/darktable.md
@@ -4,9 +4,9 @@ id: darktable
 weight: 10
 ---
 
-The `darktable` binary starts darktable with its GUI and full functionality. This is the standard way to use darktable.
+The `darktable` binary starts darktable with its GUI and full functionality. This is the standard way to use darktable. The exact command to launch the binary [depends on your OS](../../special-topics/program-invocation/invocation-per-OS.md).
 
-`darktable` can called with the following command line parameters:
+The following command line parameters are available:
 
 ```
 darktable [-d {all,act_on,cache,camctl,camsupport,common,control,
--- a/content/special-topics/program-invocation/darktable.md
+++ b/content/special-topics/program-invocation/darktable.md
@@ -4,7 +4,7 @@ id: darktable
 weight: 10
 ---
 
-The `darktable` binary starts darktable with its GUI and full functionality. This is the standard way to use darktable. The exact command to launch the binary [depends on your OS](../../special-topics/program-invocation/invocation-per-OS.md).
+The `darktable` binary starts darktable with its GUI and full functionality. This is the standard way to use darktable. The command to run the binary [depends on your OS](../../special-topics/program-invocation/invocation-per-OS.md).
 
 The following command line parameters are available:
 
--- a/content/special-topics/program-invocation/invocation-per-OS.md
+++ b/content/special-topics/program-invocation/invocation-per-OS.md
@@ -12,6 +12,7 @@ Linux
      + To run e.g. `darktable-cltest`: `flatpak run --command=darktable-cli org.darktable.Darktable --arguments-go-here`
 * AppImage: Navigate to containing folder, then `./darktable.Appimage` (or however you named the file). 
      + To run e.g. `darktable-cltest`: `./darktable.Appimage darktable-cltest`
+     + An additional option to run the various binaries included in darktable (e.g. `darktable-cltest` or `darktable-cli`) you can also use a symbolic link: Create a symbolic link to your AppImage and call it e.g. `darktable-cli`. Then, when calling the AppImage through the symbolic link with `./darktable-cli`, the code inside the package checks to see if there is a binary file inside the AppImage with a name that matches the symlink name, and if so, it will be run instead of the default one.
      
 Windows
 * Navigate to darktable's program folder: `cd "\Program Files\darktable\bin"`
@@ -22,4 +23,4 @@ macOS
 * `/Applications/darktable.app/Contents/MacOS/darktable`
 * The other binaries e.g. `darktable-cltest` are also located in this folder and can be called accordingly
 
-Command line options are in any case simply appended. 
+Command line options are in any case simply appended. (e.g. `-d common`).
--- a/content/special-topics/program-invocation/invocation-per-OS.md
+++ b/content/special-topics/program-invocation/invocation-per-OS.md
@@ -1,0 +1,23 @@
+---
+title: invocation per OS
+id: invocation-per-os
+weight: 5
+---
+
+`darktable` and the other binaries (e.g. `darktable-cltest`) can be called from the console as follows:
+
+Linux
+* Installed as package or snap: `darktable`
+* Flatpak: `flatpak run org.darktable.Darktable`. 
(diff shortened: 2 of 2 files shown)


content/special-topics/program-invocation/invocation-per-OS.md
```diff
@@ -0,0 +1,25 @@
+---
```
Phemisters, line 1: I appreciate all the work you are putting in to the docs. 

To be honest I'm not sure how necessary this page is. I'd guess most users simply run from their OS application launcher. There is nothing unusual about how darktable, etc are run from the command line. We have enough difficulty describing darktable features without trying to educate people on how to use their OS.

Other people may feel differently.

Phemisters (commented):
I don't mean to be unhelpful.

content/special-topics/program-invocation/invocation-per-OS.md
anoderay: Thanks for your take. You are of course right that we don't need to explain their OS to people. 

I still think this is worthwhile having. With the flatpak there isn't a directly accessible *binary* to apply CLI-options to (like .e.g running with `-d common`) which the [mentiones](https://docs.darktable.org/usermanual/development/en/special-topics/program-invocation/darktable/). Also if one uses the appimage information on how to run e.g. darktable-cltest can be found on pixls if you google for it but in my opinion this is useful to have in the docs. 

@victoryforce what is your opinion on this?

anoderay: @Phemisters not much input coming in from others..would you be OK with merging this? just today I read up in my own documentation here on how to use dt-cltest with the appimage. Sure the target audience for this information is slim but that can be said about a lot of pages in the docs.

victoryforce: I have mixed feelings about this. It is probably worth explaining the specifics of launching AppImage and Flatpak. Moreover, regarding the darktable bundle in AppImage, here I made a custom launch script that allows you to launch another packaged application in one of two ways: either by specifying the name of the application as the first argument, or by creating a symlink to the bundle with the name of this application and launching it through the symlink. This really needs to be documented.

anoderay: @victoryforce 
"here I made a custom launch script that allows you to launch another packaged application" 
are you referring to what makes `./darktable.Appimage darktable-cltest` possible?

I unfortunately don't completely understand what you mean by setting and launching through the symlink.

victoryforce: > @victoryforce "here I made a custom launch script that allows you to launch another packaged application" are you referring to what makes `./darktable.Appimage darktable-cltest` possible?

Yes. See https://github.com/darktable-org/darktable/pull/17995

> I unfortunately don't completely understand what you mean by setting and launching through the symlink.

For example, let's say our package is called darktable.AppImage. We create a symbolic link to this file and call it darktable-cli. Then, when we run this AppImage by calling ./darktable-cli, the code inside the package checks to see if there is a binary file inside the AppImage with a name that matches the symlink name, and if so, it will be run instead of the default one.

This was introduced in https://github.com/darktable-org/darktable/pull/17841

anoderay: @victoryforce neat, I added it. I still think this artice is worthwhile having. Especially because when we document the symbolic-link-invocation we, in my opinion, should cover the whole range.

content/special-topics/program-invocation/darktable.md
```diff
@@ -4,9 +4,9 @@ id: darktable
 weight: 10
 ---
 
-The `darktable` binary starts darktable with its GUI and full functionality. This is the standard way to use darktable.
+The `darktable` binary starts darktable with its GUI and full functionality. This is the standard way to use darktable. The exact command to launch the binary [depends on your OS](../../special-topics/program-invocation/invocation-per-OS.md).
```
Phemisters: I would consider repacing `The exact command to launch the binary...` with `The command to execute darktable...` or `The command to run darktable...`. In this change we are using `call`, `run`, `execute`, and `launch` to mean the same thing.

content/special-topics/program-invocation/invocation-per-OS.md
```diff
@@ -0,0 +1,28 @@
...
+title: invocation per OS
+id: invocation-per-os
+weight: 5
+---
+
+`darktable` and the other binaries (e.g. `darktable-cltest`) can be called from the console as follows:
```
Phemisters: I would suggest replacing `called` with `executed` or `run`.

content/special-topics/program-invocation/invocation-per-OS.md
```diff
@@ -0,0 +1,28 @@
...
+
+Linux
+* Installed as package or snap: `darktable`
+* Flatpak: `flatpak run org.darktable.Darktable`. 
+     + To run e.g. `darktable-cltest`: `flatpak run --command=darktable-cli org.darktable.Darktable --arguments-go-here`
```
Phemisters: Are you sure that this command runs `darktable-cltest` - I have not tried it but it looks like it would run `darktable-cli`.

content/special-topics/program-invocation/invocation-per-OS.md
```diff
@@ -0,0 +1,28 @@
...
+* AppImage: Navigate to containing folder, then `./darktable.Appimage` (or however you named the file). 
+     + To run e.g. `darktable-cltest`: `./darktable.Appimage darktable-cltest`
+     + An additional option to run the various binaries included in darktable (e.g. `darktable-cltest` or `darktable-cli`) is through a symbolic link: 
+          * Create a symbolic link to your AppImage and call it e.g. `darktable-cli`. 
+          * Then, when calling the AppImage through the symbolic link with `./darktable-cli`, the code inside the package checks to see if there is a binary file inside the AppImage with a name that matches the symlink name, and if so, it will be run instead of the default one.
+     
+Windows
+* Navigate to darktable's program folder: `cd "\Program Files\darktable\bin"`
+* Then `.\darktable.exe`
```
Phemisters: In some bulleted lists there is a full stop at the end of each list item, and sometimes there is not. Perhaps be consistent about it, unless there is a reason not to be.

Phemisters (approved):
I'm happy to merge this - I've left a few minor comments to consider. Let me know when you are ready for it to be merged.

content/special-topics/program-invocation/darktable.md
anoderay: Would you prefer to unify all the the "execution" verbs to the same?

content/special-topics/program-invocation/invocation-per-OS.md
anoderay: error on my side copy-pasting from pixls...

content/special-topics/program-invocation/invocation-per-OS.md
anoderay: I went over the document again with this in mind. The [general consensus](https://apastyle.apa.org/style-grammar-guidelines/lists/bulleted) seems to be to use punctuation on complete sentences, and don't on words, fragments etc.

PR #874, opened 2026-05-17
Add doc for welcome pages

Phemisters:
Relates to darktable PR: https://github.com/darktable-org/darktable/pull/20456

Add basic documentation for welcome pages and the associated preference.

I didn't document each individual page because:
1. they might change and we'd have to keep updating
2. they are for beginners, and if they need the manual to explain the welcome pages, then the welcome pages are not really helping as much as we'd like

Also, not really sure where in the manual to place this page - open to suggestions.


content/overview/welcome.md
```diff
@@ -10,5 +10,4 @@ On the first run of darktable you will be presented with some welcome pages, to
 
 Use the 'prev' and 'next' buttons to move through the pages, and use 'close' to exit the welcome pages at any time.
```
anoderay: I think this line can be omitted. Looks fine to me otherwise.

anoderay (commented):
I'd put this in "preferences and settings" after "overview". It is ultimately for setting preferences.

Phemisters:
> I'd put this in "preferences and settings" after "overview". It is ultimately for setting preferences.

I put it early in the manual because it is the first thing they see when they run the app for the first time. I don't see much value mentioning it in the preferences - I doubt anyone would read it there.

I was thinking a new user might look at the manual before deciding to try the app, and they would think _oh good, there is some assistance in setting up_. Once they have run the app, the welcome pages are pretty much self-explanatory.

So I've removed the page, and just left the preference setting.

What do you think?

anoderay:
> So I've removed the page, and just left the preference setting.
Sounds correct to me.

PR #892, opened 2026-05-27
rename drawn masks for visibility of AI masks

anoderay:
Going through the docs I noticed that with the current setup AI masks are kinda hidden and can't be easily discovered though the search function. So I suggest renaming the "drawn masks" to "drawn masks & AI masking". An alternative could be to remove the AI masks from the drawn masks and give them their own page within the masking section.

### DIFF
--- a/content/darkroom/masking-and-blending/masks/drawn.md
+++ b/content/darkroom/masking-and-blending/masks/drawn.md
@@ -1,5 +1,5 @@
 ---
-title: drawn masks
+title: drawn masks & AI masking
 id: drawn
 weight: 20
 ---


andriiryzhkov:
@anoderay : I agree with you that AI masks are kinda hidden now. But I would **advise against generalizing** AI masking under the drawn masks. Let me explain what I mean. For the next development cycle I have plans for depth masks. That would also be "AI mask" but of different type, and definitely not a drawn mask. Most probably it will be a new kind of parametric mask. So if you add "AI masking" to drawn masks now, we will need to change it somehow in a half of year.

If you want to highlight AI masking capabilities in the documentation, I may suggest adding a separate new page with a short summary and link(s) to specific page section where it is documented now.

victoryforce:
Just a note on style, as I have no opinion on whether it's better to add a separate page or not. Ampersands should only be used in some special contexts. Outside of those, it's considered lazy writing. In formal writing, you should use "and".

 See https://bernoff.com/blog/why-to-avoid-ampersands-in-formal-writing

anoderay:
> If you want to highlight AI masking capabilities in the documentation, I may suggest adding a separate new page with a short summary and link(s) to specific page section where it is documented now.

@andriiryzhkov sounds like a good solution. I'll get to it


> Just a note on style, as I have no opinion on whether it's better to add a separate page or not. Ampersands should only be used in some special contexts. Outside of those, it's considered lazy writing. In formal writing, you should use "and".
> See https://bernoff.com/blog/why-to-avoid-ampersands-in-formal-writing

@victoryforce Thanks for pointing that out. I didn't know that & is named Ampersand before. I followed the styling already present in the docs:

grafik

And also reading the article you quoted it seems fine to me to use the Ampersand in a title (even though that case is not explicitly named). IMHO it improves readability by visually bisecting the title in the two entities.

PR #898, opened 2026-05-29
Rename "object" mask to "AI object" in drawn-masks docs

andriiryzhkov:
I just realised that in the darktable UI the AI-generated mask shape is called **"AI object"**, but in the docs (`content/darkroom/masking-and-blending/masks/drawn.md`) it was written as just "object". Renamed both occurrences so the docs match the UI.


Phemisters:
Good catch.

PR #904, opened 2026-06-09
Add documentation on how dtdocs is set up and deployed

anoderay:
I wanted to start documenting how dtdocs is setup and maintained/deployed. This is to be considered a first draft and could probably be expanded upon. Looking for feedback/comments @Phemisters .


content/contributing/setup-of-dtdocs.md
```diff
@@ -0,0 +1,9 @@
+---
+title: setup of the docs
```
Phemisters: I wonder if it would be better to call it something like _dtdocs infrastructure_ to help make it clear the typical user does not need to know this?

content/contributing/setup-of-dtdocs.md
```diff
@@ -0,0 +1,9 @@
...
+id: setup-of-dtdocs
+weight: 2
+---
+
+darktable's docs are maintained through their [GitHub repository](https://github.com/darktable-org/dtdocs). A GitHub action is set up to run `tools/build-all.sh` which will generate html, epub and pdf outputs using Hugo. Those are saved to the repository's branch [gh-pages](https://github.com/darktable-org/dtdocs/tree/gh-pages) and deployed through GitHub pages to https://darktable-org.github.io/dtdocs/.
```
Phemisters: I wonder if it would be worth mentioning that the dtdocs deployment does not generate translations, but that  the pixls.us deployment does?

Phemisters (commented):
I think it is indeed a good idea to have somewhere to capture information about how things are built and deployed. A couple of small suggestions, but feel free to ignore  them.

content/contributing/setup-of-dtdocs.md
anoderay: what do you think of "development and deployment"? See commit.

content/contributing/setup-of-dtdocs.md
anoderay: Didnt know that - thanks!

content/contributing/setup-of-dtdocs.md
Phemisters: Sounds good.

content/contributing/development-deployment.md
```diff
@@ -0,0 +1,18 @@
...
+ 
+## Official deployment to darktable.org
+ 
+The official home of the docs is the pixls.us server infrastructure at https://docs.darktable.org/usermanual/development/. There, the repository is pulled regularly and the documentation is built with Hugo, including translations as well as EPUB and PDF outputs. The Build Log can be found [here](https://docs.darktable.org/buildlog.txt).
+
+## Versioning of the docs 
+Up to 4.6 versioned copies of the docs were maintained (e.g. [../usermanual/4.6/](https://docs.darktable.org/usermanual/4.6/)). From there on only one version documenting the current development version is maintained ([../usermanual/development/](https://docs.darktable.org/usermanual/development/)). Links to the released versions following 4.6 are linked to the development version (e.g. [../usermanual/5.4/](https://docs.darktable.org/usermanual/5.4/) links to [../usermanual/development/](https://docs.darktable.org/usermanual/development/)).
```
Phemisters: Two dots in a file path generally means the parent directory, but here I think the dots indicate that only part of the path is included. I think using three dots for this would be clearer.

Phemisters (approved):
Looks good. I've left one minor comment, after which I'm happy to merge.

content/contributing/development-deployment.md
anoderay: done

Issue #132, opened 2020-11-18
Include detail from header tooltips in module reference


elstoc:
Processing module headers in darktable now contain a brief summary of the module along with details of the purpose, input, output and working color spaces. 

Consider including this in the dtdocs module reference sections. Some of it is already included but not yet visible.

Should we use the yaml metadata to include this detail or include it as separate content?

paperdigits:
Can you provide some samples of what these tooltips look like?

elstoc:
![Screenshot_2020-11-19_07-52-39](https://user-images.githubusercontent.com/9555491/99637105-47722780-2a3c-11eb-95e9-e9b67aa43791.png)

raublekick:
If there's still interest in this I'm happy to grind through the modules. I did a little proof of concept. For my own possible future reference: It's "easy" to find the info in the darktable codebase searching for instances of `dt_iop_set_description`.  You can use the `name` function to see the name of the module to match it to a .md file in the docs:

```c
const char *name()
{
  return _("AgX");
}

const char **description(dt_iop_module_t *self)
{
  return dt_iop_set_description(self,
                                _("applies a tone mapping curve.\n"
                                    "inspired by Blender's AgX tone mapper"),
                                _("corrective and creative"),
                                _("linear, RGB, scene-referred"),
                                _("non-linear, RGB"),
                                _("linear, RGB, display-referred"));
}
```
Added to the AgX yaml:
```yaml
---
title: AgX
id: agx
weight: 10
params:
    module_info:
        description: "applies a tone mapping curve. inspired by Blender's AgX tone mapper"
        purpose: "corrective and creative"
        input: "linear, RGB, scene-referred"
        processing: "non-linear, RGB"
        output: "linear, RGB, display-referred"
---
```

Rendered:

<img width="1134" height="561" alt="Image" src="https://github.com/user-attachments/assets/38b74086-66b6-4cd8-89f3-97d5b10ab688" />

I had to edit the theme template but I didn't bother trying to style it any particular way.

elstoc:
On reflection I'd prefer to have this as actual markdown rather than interpreted frontmatter. The more generally portable we can keep the markdown the better, just in case we wanted to move away from Hugo in the future.

Also, it's kinda techie info so might be better at the end of the documentation?

Honestly though, I go back and forth on whether or not this is a good idea at all

raublekick:
[...]

PR #910, opened 2026-06-22
Add the individual processing module tooltip information

Phemisters:
Relates to dtdocs issue: #132 

In the processing module source code of each module, there is a call to `dt_iop_set_description` which sets the information shown in the tooltip when the module name is hovered over.

I have added this information to each module's page using a [Hugo _details_ shortcode](https://gohugo.io/shortcodes/details/).

* In the html it is presented as a collapsible `details` tag.
* In the PDF it is show in a box to make it clear it is to be read as grouped information.
* In the epub it is again shown as a collapsible details section.

I have added a note in the module overview trying to describe it.

I'm open to suggestions of how best to format it, and how best to describe it in the overview.

Some deprecated modules don't have a `dt_iop_set_description` call, but I have not specifically highlighted that (as people as less likely to be looking at the pages of deprecated modules anyway).

### DIFF
--- a/content/module-reference/overview.md
+++ b/content/module-reference/overview.md
@@ -9,7 +9,7 @@ The modules in this reference section are broken down into two distinct types:
 [processing modules](./processing-modules/_index.md)
 : Processing modules are used exclusively in the darkroom view. Each module performs a processing operation on the image before passing its output to the next module for further processing. Together this sequence of processing steps forms the [pixelpipe](../darkroom/pixelpipe/the-pixelpipe-and-module-order.md).
 
-  Processing modules have a _Synopsis_ section which gives:
+  Processing modules have a _Synopsis_ section which gives a brief summary of the module (the information can also be seen as a tooltip when hovering over the module name). The following information is shown:
     * a brief outline of the purpose of the module
     * an indication of the purpose (e.g. corrective or creative)
     * the processing color space (e.g. non-linear, Lab)
--- a/content/module-reference/overview.md
+++ b/content/module-reference/overview.md
@@ -9,6 +9,14 @@ The modules in this reference section are broken down into two distinct types:
 [processing modules](./processing-modules/_index.md)
 : Processing modules are used exclusively in the darkroom view. Each module performs a processing operation on the image before passing its output to the next module for further processing. Together this sequence of processing steps forms the [pixelpipe](../darkroom/pixelpipe/the-pixelpipe-and-module-order.md).
 
+  Processing modules have a _Synopsis_ section which gives:
+    * a brief outline of the purpose of the module
+    * an indication of the purpose (e.g. corrective or creative)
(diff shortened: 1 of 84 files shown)


anoderay:
Neat Idea. I am not too sure about hiding this behind a drop down but have to dwell on that. Maybe replace "synopsis" with "technical information"? As it is fairly technical?

Phemisters:
> Maybe replace "synopsis" with "technical information"?

I've pushed a change to use _technical information_.

anoderay:
Thanks. I suppose you somehow automated pulling these strings from the codebase, right? I wonder how we will deal with translating these. If they get manually translated it will be a lot of work and discrepancies with the strings used in dt might come up.

Are you able to extract the translated strings, too? I might have to look into how we could automate pushing those into the translations.

Phemisters:
> Are you able to extract the translated strings, too? I might have to look into how we could automate pushing those into the translations.

I did use a custom tool to do this initial update of all the files. It takes the data from code like this

```c
 return dt_iop_set_description(self,
                                _("applies a tone mapping curve.\n"
                                    "inspired by Blender's AgX tone mapper"),
                                _("corrective and creative"),
                                _("linear, RGB, scene-referred"),
                                _("non-linear, RGB"),
                                _("linear, RGB, display-referred"));

```

so I suppose ` _("corrective and creative"),` is a translatable string in darktable. It would certainly be possible to compile a list of all the unique strings used, is that what you were thinking?

anoderay:
@Phemisters yeah something like that, though the translated strings are in PO/POT Files in darktable anyway; so they should already be connected to the original strings. I will try to get those into weblate to make work easier for translators.

Phemisters:
> It would certainly be possible to compile a list of all the unique strings used, is that what you were thinking?

Although I can still do that, I see that the po files contain info like this;

```
#. type: Plain text
#: content/module-reference/processing-modules/channel-mixer.md
#: content/module-reference/processing-modules/rotate-perspective.md
#: content/module-reference/processing-modules/crop.md
#: content/module-reference/processing-modules/rgb-primaries.md
#: content/module-reference/processing-modules/lut-3D.md
#: content/module-reference/processing-modules/color-calibration.md
#: content/module-reference/processing-modules/color-look-up-table.md
#: content/module-reference/processing-modules/color-balance-rgb.md
#: content/module-reference/processing-modules/color-balance.md
#: content/module-reference/processing-modules/color-correction.md
#: content/module-reference/processing-modules/crop-rotate.md
#: content/module-reference/processing-modules/external-raster.md
#, no-wrap
msgid ""
"purpose\n"
": corrective or creative.\n"
msgstr ""

```

which to me looks like it already de-duplicates the strings. Is it still a concern?

anoderay:
@Phemisters I'll look into it. I'll ping you when I know more :-) .

PR #913, opened 2026-06-23
updated workflow with hugo

anoderay:
Added instructions for running hugo locally to preview changes.

### DIFF
--- a/content/contributing/workflow.md
+++ b/content/contributing/workflow.md
@@ -69,6 +69,9 @@ Alternatively you can sync your fork and local copy via terminal:
 10. If you need to make further corrections using the current branch, simply repeat 3-8. For the command in step 3 you can omit `-b` and in step 8 you can just do `git push`.
 11. After you are done and your pull request is merged or you are finished with your work for this specific pull request, execute `git checkout master` to return to the master branch. I recommend working on only one branch/pr at a time if you don't know how git works.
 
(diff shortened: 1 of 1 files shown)


content/contributing/workflow.md
```diff
@@ -69,6 +69,9 @@ Alternatively you can sync your fork and local copy via terminal:
 10. If you need to make further corrections using the current branch, simply repeat 3-8. For the command in step 3 you can omit `-b` and in step 8 you can just do `git push`.
 11. After you are done and your pull request is merged or you are finished with your work for this specific pull request, execute `git checkout master` to return to the master branch. I recommend working on only one branch/pr at a time if you don't know how git works.
 
+## running `hugo` locally to see changes
```
Phemisters: I'd capitalize `running`.

The README also mentions running Hugo (and building the assets) - I wonder if that part of the README should be moved here, to avoid having the same information in two places? Just a thought.

Phemisters (approved):
Minor comments but looks good to me.

content/contributing/workflow.md
anoderay: I think it is fine to have the hugo reference in both places. The Readme is the more general reference document; the workflow.md a tutorial for newcomers. 

That being said, it could make sense to generally work through the readme to check if we want to change things.

PR #930, opened 2026-06-29
update sytles and conventions to include TOC and details

anoderay:
documents how the TOC is enabled and the block of technical information for the processing modules.

### DIFF
--- a/content/contributing/styles-conventions.md
+++ b/content/contributing/styles-conventions.md
@@ -2,6 +2,7 @@
 title: styles & conventions
 id: styles-conventions
 weight: 10
+include_toc: true
 ---
 
 The following details the writing style and documentation conventions used in this manual.
--- a/content/contributing/styles-conventions.md
+++ b/content/contributing/styles-conventions.md
@@ -91,6 +91,7 @@ For each of the processing modules there is a special (collapsible) block of tec
 
 ```
 {{</* details summary="Technical information" class="technical-info" */>}}
+
 description
 : applies a tone mapping curve. inspired by Blender's AgX tone mapper.
 
@@ -105,6 +106,7 @@ processing
 
 output
 : linear, RGB, display-referred
+
 {{</* /details */>}}
 ```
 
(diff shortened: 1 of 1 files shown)


content/contributing/styles-conventions.md
```diff
@@ -80,6 +85,31 @@ weight
...
+
+processing
+: non-linear, RGB.
+
+output
+: linear, RGB, display-referred
+{{< /details >}}
+```
+
+This should mirrror the information available as tooltip upon hovering the mouse over a processing module.
```
Phemisters: Typo: Extra `r` in mirror.

content/contributing/styles-conventions.md
```diff
@@ -80,6 +85,31 @@ weight
 - Changes to the content should be proposed via pull request or a similar mechanism
 - Your submissions will be copy-edited -- don't take it personally
 
+## technical information for processing modules
+
+For each of the processing modules there is a special (collapsible) block of technical information at the top:
+
+```
+{{< details summary="Technical information" class="technical-info" >}}
```
Phemisters: For me, this gets rendered by Hugo as the collapsible section, but I think the intention is probably to show what the user has to actually type.

This may be because I am using a different version of Hugo, but I had to make it like this, with extra `/*` and `*/` characters:

\```
{{</\* details summary="Technical information" class="technical-info" \*/>}}
...
{{</\* /details \*/>}}
\```


You might also want to say that authors can simply mention that this info needs added and that someone (me!) can add it later, since not everyone wants to have to extract that information. I will later write a script to go in the tools directory, but I don't have anything to share yet.

content/contributing/styles-conventions.md
```diff
@@ -65,6 +67,9 @@ weight
       section3.md                # weight: 10 (place section3 at the start of example-chapter)
 ```
 
+include_toc
+: Optional; enables or disables the collapsible table of contents.
```
Phemisters: Maybe say it is only ever used for large, complex pages.

Phemisters (commented):
Looks pretty good. I've left a few comments to consider.

content/contributing/styles-conventions.md
anoderay: Thanks for the technical correction. Considering that adding this block this is a single-time operation with relatively little typing I would leave it as is until your script (nice!) is available.

Issue #908, opened 2026-06-21
Refresh lua docs to reflect bundling


anoderay:
See here https://discuss.pixls.us/t/darktable-5-6-lua-scripts/58755

anoderay:
closed by https://github.com/darktable-org/dtdocs/pull/938

PR #938, opened 2026-06-30
update lua documentation to reflect bundling

anoderay:
Adresses https://github.com/darktable-org/dtdocs/issues/908 as starting from 5.6 lua scripts are delivered bundled with darktable. 

See also https://discuss.pixls.us/t/darktable-5-6-lua-scripts/58755 .

@wpferguson could you clarify whether the preference "disable lua scripts" disables ALL lua scripts including ones that might be specified in the now optional `luarc` or just the bundled ones?

### DIFF
--- a/content/lua/basic-principles.md
+++ b/content/lua/basic-principles.md
@@ -4,7 +4,7 @@ id: basic-principles
 weight: 20
 ---
 
-At startup, darktable will automatically run the Lua scripts `$DARKTABLE/share/darktable/luarc` (where `$DARKTABLE` represents the darktable installation directory) and `luarc` in the darktable [configuration directory](../preferences-settings/config-directory.md).
+At startup, darktable will automatically run the Lua scripts `$DARKTABLE/share/darktable/luarc` (where `$DARKTABLE` represents the darktable installation directory) and `luarc` in the darktable [configuration directory](../preferences-settings/config-directory.md). Starting from 5.6 `luarc` is not strictly required but will be respected if present. 
 
 This is the only time darktable will run Lua scripts by itself and can be used to load other scripts (like the script manager). These scripts can register callbacks to perform actions on various darktable events. This callback mechanism is the primary method of triggering lua actions. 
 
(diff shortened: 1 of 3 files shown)


wpferguson:
> could you clarify whether the preference "disable lua scripts" disables ALL lua scripts

All Lua scripts are disabled.  When script_manager starts it checks the disable lua scripts preference and if it's TRUE then script_manager just exits.

---

If the user installs the lua-scripts locally then those are executed instead of the bundled ones.

---

If you want to just run a local script from the user luarc, then specify it in the local luarc and hide script_manager in lighttable view.  script_manager will still be running, but it is just sitting idle waiting for a user interaction.  After the first time you wont see script_manager again unless you make it visible.

anoderay:
@wpferguson thanks for the clarification; I updated the docs and to get this off my list merged the pr. 

I would be grateful if you could take a look at the [three](https://docs.darktable.org/usermanual/development/en/preferences-settings/lua-options/) [affected](https://docs.darktable.org/usermanual/development/en/lua/installation/) [docs](https://docs.darktable.org/usermanual/development/en/lua/basic-principles/) whether it is actually accurate now.

anoderay:
closes https://github.com/darktable-org/dtdocs/issues/908

wpferguson:
@anoderay I owe you an apology.  I didn't even think about the Lua section in the user manual.  Thanks for taking care of it.

---

Some thoughts...

- In the overview section after the link to the Lua reference manual how about adding a link to darktable's Lua documentation (https://darktable-org.github.io/luadocs/)?
- Instead of pre-existing scripts maybe just call it darktable scripts and then wrap the discussion into a 5.6 and above and prior 5.6 which is most of the page.
- basic principles: luarc files: You might say that darktable starts script_manager and script_manager checks for a user luarc file and executes it if it's found.  My concern with `$DARKTABLE/share/luarc` is that user's may think they need to modify it, which they don't.
- In the keeping tidy section the luarc line should be `require "user/user-script"`.  Lua automatically appends `.lua` when it searches.
- Lua API - change the link to the online (and up to date) Lua API manual, https://darktable-org.github.io/luadocs/lua.api.manual/

PR #940, opened 2026-06-30
Readme overhaul

anoderay:
Reworked readme.md because some stuff was outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,19 @@
 # darktable Documentation
 
-This is the user documentation for [darktable](https://darktable.org). The content is authored in markdown and rendered to HTML using [Hugo](https://gohugo.io) as well as to PDF and ePub.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,54 +39,71 @@ Instead of yarn install you can use `npm install`.
...
 
 ```
 hugo
 ```
 
-The static files are now available to deploy to a webhost in the `public` directory. This is currently performed automatically when new changes are pushed to the master branch, via github Actions.
+The static files are now available to deploy to a webhost in the `public` directory. For docs.darktable.org this is automated, see *Deployment* below.
 
-## PDF
+### PDF
```
Phemisters, line 81: There is also a `./tools/build-pdf.sh ` available, I'm not sure if it is not mentioned for a reason.

Phemisters (commented):
Looks good to me. I've made one minor comment.

README.md
anoderay: You are right. I was somewhat lazily building upon the current readme.md but there is no reason not to reference the proper script.

Issue #943, opened 2026-07-01
document highlight preservation mode


anoderay:
Apparently introduced some time in the past; see https://github.com/darktable-org/darktable/commit/086ec9c3480b297324a21e4a5012759b679e92f7, https://github.com/darktable-org/darktable/pull/19347, https://github.com/darktable-org/darktable/pull/19375, see also [here](https://discuss.pixls.us/t/undocumented-feature-in-exposure-module/57293) and [here](https://discuss.pixls.us/t/darktable-5-6-0-released/58760/146?u=mino).

anoderay:
closed by https://github.com/darktable-org/dtdocs/pull/945

PR #945, opened 2026-07-01
Update exposure.md to include highlight preservation mode

anoderay:
adresses https://github.com/darktable-org/dtdocs/issues/943

For internal documentation: The mode is activated upon import by the function `_check_highlight_preservation` in [exif.cc](https://github.com/darktable-org/darktable/blob/master/src/common/exif.cc). In the code there the affected camera models, modes and exif tags are documented. 

@Phemisters do you think we should make this more verbose to include technical information on what cameras, modes and exif tags are relevant?

### DIFF
--- a/content/module-reference/processing-modules/exposure.md
+++ b/content/module-reference/processing-modules/exposure.md
@@ -41,6 +41,9 @@ mode
 compensate camera exposure (manual mode)
 : Automatically remove the camera exposure bias (taken from the image's Exif data).
 
+highlight preservation mode (manual mode)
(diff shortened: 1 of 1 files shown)


Phemisters:
> do you think we should make this more verbose to include technical information on what cameras, modes and exif tags are relevant?

I think that might be hard to maintain, so for now I think we should stick with what you have added. I'm happy to merge as it is now, unless you want more time to consider.

Edited: For the record, the darktable PR now has `documenation: pending`: https://github.com/darktable-org/darktable/pull/19347

PR #946, opened 2026-07-01
Add script for extracting tech info

Phemisters:
This is the first attempt at a general script to extract the _Technical information_ summary from a darktable source file, based on the _dt_iop_set_description_ call in the source file.

The intention is to have a way of standardizing the formatting.

The script is given a single arg (the source file)  and the summary is given on stdout, from where it can be copied and pasted to the dtdocs md file.

It relies on `awk` being installed, but I think it is fairly common.

Can be used by individual authors, or the summary can be added as part of the copy-editing stage.

### DIFF
--- a/tools/extract-tech-info.sh
+++ b/tools/extract-tech-info.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+#
+# Extract dt_iop_set_description technical info from a source file,
+# and format it for a dtdocs md file.
+#
+# ##################
+
+# Specify the awk program inside the script
+# Start awk program
+read -r -d '' AWK_SCRIPT <<'EOF'
+# An awk script to extract technical info summary from a source file
+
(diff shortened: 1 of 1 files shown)


anoderay:
nice - works well! Could you document the usage in a comment at the top of the script?

Phemisters:
> Could you document the usage in a comment at the top of the script?

I've added usage info to the comment at the start.

tools/extract-tech-info.sh
```diff
@@ -1,7 +1,21 @@
 #!/usr/bin/env bash
 #
-# Extract dt_iop_set_description technical info from a source file,
-# and format it for a dtdocs md file.
+# Extract the dt_iop_set_description technical info from a darktable processing
+# module source file, and format it for a dtdocs md file.
+#
+# Uasge:
```
anoderay: typo, feel free to merge when fixed :-)

anoderay (changes requested):

PR #948, opened 2026-07-02
addded punctuation in technical information

anoderay:
I suppose the omission of the last period which I added here was by accident. Stumbled upon it while trying to automate translations for these blocks. 

@Phemisters please take note for your scripting :-) .


Phemisters:
Thanks.

PR #950, opened 2026-07-02
Added script used to seed translations of "technical information" blocks

anoderay:
Ingests the po files of darktables repo itself, the POT file of dtdocs and generates PO files containing translations of the strings used in "technical information" which were then uploaded to weblate directly to seed the languages with the translations.


Phemisters:
This looks like a great tool. I ran it once and the outputs look okay to me.

PR #964, opened 2026-07-08
feat: add version switcher dropdown and 404 fallback page

fabiosirna:
### Description of the Feature
Currently, navigating between different versions of the darktable user manual is difficult for users, as there is no clear UI element to jump between older releases and the development version. 

This PR introduces a **Version Switcher** and a **Fallback 404 Page** to improve the overall UX.

### What has been done:
1. **Version Switcher Dropdown:** Added a native Bootstrap dropdown in the top navbar (`page-header-navbar-large.html`). It automatically calculates the relative path of the current page, allowing the user to seamlessly switch to the same page in versions `4.8`, `4.6`, `4.4`, and `Development (Master)`.
2. **Helpful 404 Fallback:** If a user uses the dropdown to jump to an older version but the current page didn't exist back then, they would hit a dead end. I have updated/created a clean `404.html` template. Now, instead of a broken layout, users get a friendly message redirecting them to `https://www.darktable.org/resources/`.

### Note on SEO (Future Work)
This PR is targeted at the `master` (development) branch, so no SEO changes were made here. However, to prevent older versions of the manual from outranking the current development/stable versions on Google, I suggest opening separate PRs on the historic branches (e.g., `release-4.6`) to include a `<meta name="robots" content="noindex, follow">` tag in their `head-meta.html`. I can help with this in future PRs if the team agrees.


anoderay:
Thanks! I think we will have to update all branches as the template into which you (I assume) added the version switcher is not shared between branches as far as I know and would thus become a one-way trip ;-) .

Might also be worth thinking about adding the theme(s) as an own repository embedded as a git submodule. It was this way in the past but I don't know why the previous maintainers changed it.

content/module-reference/processing-modules/color-calibration.md
```diff
@@ -282,6 +282,8 @@ Note that the target value is not reset when you reset the module itself, but is
...
 
+When selecting the area to measure chromaticity for color calibration in area mode, it is crucial to note that the selected area remains persistent as you edit subsequent images. To define the area, click on the eyedropper icon located to the right of the module. This icon displays the tooltip "set white balance to detected from area" on mouseover. Then, left-click and drag the mouse cursor over the desired region of the image to select the measurement area. This selection ensures consistency in your calibration workflow, as the same target area will be used for measuring chromaticity across multiple images, unless explicitly reset. To reset the selected area, simply right-click on the image. This feature allows you to maintain a consistent calibration reference, enhancing accuracy and efficiency during batch editing or sequential adjustments.
```
Phemisters: I think this change should really be a separate PR.

themes/hugo-darktable-docs-theme/layouts/partials/page-header-navbar-large.html
```diff
@@ -9,7 +9,19 @@
...
+               </a>
+               <div class="dropdown-menu dropdonw-trans" role="listbox">
+                  {{ $cleanPath := replace .RelPermalink "/usermanual/development/" "/" }}
+                  
+                  <a class="dropdown-item" href="/usermanual/development{{ $cleanPath }}">Development (Master)</a>
+                  <a class="dropdown-item" href="/usermanual/4.8{{ $cleanPath }}">Version 4.8</a>
+                  <a class="dropdown-item" href="/usermanual/4.6{{ $cleanPath }}">Version 4.6</a>
+                  <a class="dropdown-item" href="/usermanual/4.4{{ $cleanPath }}">Version 4.4</a>
+               </div>
+            </li>
```
Phemisters: I'm a little confused - there is still a `</li>` closing tag at line 36, which looks like it does not have a matching `<li>`.  Is it no longer needed?

Phemisters (commented):
This is looking good. I'll do some more testing, but I've got a couple of minor comments for now.

themes/hugo-darktable-docs-theme/layouts/partials/page-header-navbar-large.html
fabiosirna: Good catch! You're totally right, that was a leftover from my edits. I've just pushed a fix to correct the HTML structure and restore the language dropdown layout. Thanks for spotting it!

content/module-reference/processing-modules/color-calibration.md
fabiosirna: Oops! This file was modified locally during some previous tests and accidentally got mixed into my commits. I didn't mean to bundle it with the UX/UI changes.

I have just removed it and pushed a clean update, so it should be completely gone from this PR now. Sorry about this.

fabiosirna:
> Thanks! I think we will have to update all branches as the template into which you (I assume) added the version switcher is not shared between branches as far as I know and would thus become a one-way trip ;-) .
> 
> Might also be worth thinking about adding the theme(s) as an own repository embedded as a git submodule. It was this way in the past but I don't know why the previous maintainers changed it.

You are completely right, @anoderay. Since the theme templates are independent on each branch, the changes won't automatically propagate.

Since we already planned to add the noindex tag to all those older branches, I think the cleanest approach is for me to bundle both the version switcher HTML and the noindex meta tag into the upcoming PRs for each legacy branch. That way, we can ensure every single version gets properly updated and tested without relying on cherry-picking, which might trigger template conflicts.

For now, I'll wait for your green light on this PR, and then I'll start prepping the combined (UI + SEO) PRs for the older branches!

Phemisters:
I don't have any other comments. I don't have any views on the ideas to split themes to a separate repository, and I don't know how we'd deploy some of the changes.

anoderay:
Alright I checked this out as well now - looks fine to me. Concerning how we will deal with theming in the future i have opened a [new issue](https://github.com/darktable-org/dtdocs/issues/973). 

Merged! @fabiosirna go ahead with the now following work :-) -

PR #976, opened 2026-07-14
[3.6] fix infinite loading because of dead analytics

anoderay:
Backport of #869 to the release-3.6 branch.

Analytics ran on a piwik instance which is (presumably permanently) offline. See #868.

This removes analytics.html and the corresponding reference in the layout for this release branch, which was still affected since it wasn't caught by the fix on master.


anoderay:
@Phemisters this one should be fine to merge as 3.6 doesn't have the submodule setup.

Issue #992, opened 2026-07-16
Use of deprecated modules


lebmich:
Modules _crop and rotate_  and _spot removal_ are mentioned in section _importing sidecar files generated by other applications_, but they are deprecated and should no longer be used.

Phemisters:
@lebmich Thanks for raising this. I don't have access to a lightroom XMP file so I can't update this to say what darktable currently does for these. Do you have a lightroom sample which has cropping, rotation, and cloning / spot removal? Do you know what darktable does with those changes?

lebmich:
@Phemisters Sorry! But I've never used Lightroom. I can't help you. As a result, darktable is no longer able to handle Lightroom's "crop and rotate" and "spot removal" development steps.

Phemisters:
I am not a darktable dev, so what I am about to say might be wrong. @lebmich - please correct me if I'm wrong.

Looking at the code in [lightroom.c](https://github.com/darktable-org/darktable/blob/be7259b546e75c0afabcca8f9186fba9156b89ce/src/develop/lightroom.c#L1385) it looks to me like the import still uses the deprecated module (_clipping_ is, I think, _crop and rotate_). This is sort of okay - people can still edit images which use deprecated modules, they just can't use new instances of them.

If there is a problem, it looks to me like it is in darktable (which should presumably not be using deprecated modules in the import), not the doc.

Phemisters:
I see there is an open darktable issue (https://github.com/darktable-org/darktable/issues/16200) about this, and I don't think it is likely to change soon. I'll add a note to the docs.

anoderay:
Closed by https://github.com/darktable-org/dtdocs/pull/995

PR #995, opened 2026-07-18
Mention use of deprecated in Lightroom XMP import

Phemisters:
Related to dtdocs issue: #992 

The issue points out that the import of Lightroom-generated XMP files uses deprecated modules. This is discussed in the darktable issue: https://github.com/darktable-org/darktable/issues/16200

I think it is worth adding a note about this.


anoderay:
Closes https://github.com/darktable-org/dtdocs/issues/992

PR #997, opened 2026-07-20
Update midi-device-support.md to include Akai LPD8

anoderay:
@ClintonDavis PR as proposed by you :-) . How do the absolute encoders work when switching images? That may be worth a comment too.


Phemisters:
This looks good to me. I'll not merge yet in case @ClintonDavis has any comments.

ClintonDavis:
Thanks, would it be ok to hold off merge for a couple of days? 

After writing this, I had a couple of program crashes, so I'd like to see if that continues as I do some work this week and if so will change the text to reflect it.

I'll also try the encoders for switching images (I suspect it won't go well because using them for zooming in and out on the lighttable was sketchy). I am using the pressure pads for switching images in both the lighttable and darkroom which is working well.

Phemisters:
> Thanks, would it be ok to hold off merge for a couple of days?

@ClintonDavis do you have any update on this?

Issue #924, opened 2026-06-26
update filmic documentation to remove references to basecurve etc.


anoderay:
Filmics docs include a number comparisons to basecurve etc.; I feel like these are no longer needed and could be cleaned up as newer user probably dont know darktables older display referred paradigm and associated modules.

victoryforce:
Direct curve manipulation tools are exactly what users who have come to darktable from other programs are familiar with. Don't take the comparison with such tools as a reference to something outdated and deprecated in darktable. Think of it as something from the user's previous experience.

anoderay:
That makes sense but the docs should then reference these things properly. In the current state I personally would read filmics docs and be confused by all other modules being referenced which I then would probably not need and never meet again in the current beginner tutorials going around ;-). Filmics docs could generally do with an [overhaul](https://discuss.pixls.us/t/filmic-rgb-call-to-overhaul-the-docs/58881) in my opinion.

anoderay:
Addressed by https://github.com/darktable-org/dtdocs/pull/999

PR #999, opened 2026-07-21
restructure and update filmic rgb article

anoderay:
I overhauled filmic rgbs documentation with help by claude.ai Opus 4.8. I don't know how I feel myself about the heavy AI involvement in this process but it made for a much more readable, shorter (-28%) article in my opinion. 

Key Points:

__Structure__
- `prerequisites` > `before you start`
- New `legacy color science` section: everything concerning v3–v6 is collected here. The main body now describes v7 only
- `caveats` > `troubleshooting`
- `background` reduced to the v6/v7 logic

__Content__
- Removed all four references to *base curve*; *sigmoid* and *AgX* are mentioned
- Less dogmatic tone
- Adopted in substance from the Ansel documentation: the rule of thumb for the linear region, the order in which to check colorless highlights, and decision guidance for *contrast in shadows/highlights*

__Corrections verified against `filmicrgb.c`__
- I had claude read filmicrgb.c to check the article for factual correctness
- Since spline v3, contrast scales with the dynamic range automatically (`slope = contrast * DR / 8`). 
- "A contrast of 1 disables the S-curve" was wrong: 1.0 is the default and produces a slope of 1.5 at 12 EV
- The linear region is a percentage of the display range, not of the scene dynamic range
- The desaturation-free linear region applies to v3-v5 only (`filmic_desaturate_v4` does not use it) > moved to the legacy section
- Added: `RGB euclidean norm (legacy)`, `uniform` for the noise distribution, and v7 in the list of selectable versions
- GUI strings adopted, without hyphens ("middle gray luminance", "auto adjust hardness")

Addresses https://github.com/darktable-org/dtdocs/issues/924

### DIFF
--- a/content/module-reference/processing-modules/filmic-rgb.md
+++ b/content/module-reference/processing-modules/filmic-rgb.md
@@ -1,359 +1,335 @@
+
+
+Filmic rgb · MD
 ---
 title: filmic rgb
 id: filmic-rgb
 weight: 10
 include_toc: true
 ---
-
+ 
 {{< details summary="Technical information" class="technical-info" >}}
-
+ 
 description
 : apply a view transform to prepare the scene-referred pipeline for display on SDR screens and paper prints while preventing clipping in non-destructive ways.
-
+ 
 purpose
 : corrective and creative.
-
+ 
 input
 : linear or non-linear, RGB, scene-referred.
-
+ 
 processing
 : non-linear, RGB.
-
+ 
 output
 : non-linear, RGB, display-referred.
-
+ 
 {{< /details >}}
-
+ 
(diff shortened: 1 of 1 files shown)


content/module-reference/processing-modules/filmic-rgb.md
```diff
@@ -1,359 +1,335 @@
+
+
+Filmic rgb · MD
```
Phemisters: The extra 3 lines added at the top of the file should probably be removed.

Phemisters (commented):
I'm not a big user of filmic rgb, so can't comment much on the detail. To me, this looks good. I've left one minor comment for consideration, after which I'be okay with merging this.

PR #1000, opened 2026-07-22
Exif editor note for added tags

Phemisters:
Related to darktable PR: https://github.com/darktable-org/darktable/pull/20872

Adds a note to the metadata editor about when the added exif tags are read.


anoderay:
Thanks!

PR #1009, opened 2026-07-24
Added statement on importing preset files from an external source.

ratcheer:
Thank you for taking the time to help the darktable documentation!

Adding a small paragraph to Managing Presets to mention the ability to import external preset files to your darktable installation.

### DIFF
--- a/content/darkroom/processing-modules/presets.md
+++ b/content/darkroom/processing-modules/presets.md
@@ -96,6 +96,8 @@ format
 
 Both user-defined and internal presets can be viewed and managed from within [preferences > presets](../../../preferences-settings/presets.md).
 
+If you get a preset file from an external source, such as downloaded from another user, use the Import button at the bottom right of the presets preferences screen.
+
 ---
 
 **Note:** If you create a user-defined preset with the same name as a built-in preset, your preset will override the built-in version, which will no longer be accessible.


content/darkroom/processing-modules/presets.md
```diff
@@ -96,6 +96,8 @@ format
 
 Both user-defined and internal presets can be viewed and managed from within [preferences > presets](../../../preferences-settings/presets.md).
 
+If you get a preset file from an external source, such as downloaded from another user, use the Import button at the bottom right of the presets preferences screen.
```
anoderay, line 99: I'd suggest "To add an external preset file, e.g. provided by another user, use the import [...]"

anoderay (changes requested):
Thanks for your contribution Tim, minor review (if you edit your local branch and git-push it will automatically update this PR, too).

Phemisters:
Thank you for contributing to the docs. The sentence before your addition includes a link to the preferences -> presets page, and that doc page already mentions the import option. Having multiple pages with the same information tends to increase future maintenance problems.

As an alternative, I wonder if you'd consider changing the `presets.md` page where it says:

> User-defined presets can be imported from exported `.dtpreset` files using the “import” button at the bottom of the screen. 

to maybe highlight that this file could be from an external source?

Edited: I see that @anoderay was commenting at about the same time, sorry about that :smile:

ratcheer:
I will try. It may be tomorrow before I can get to it.

But when I was looking for this information in the documentation, I never
saw the sentence you refer to. It is apparently hidden by an indirection.
Thank you for bringing this to my attention.

Phemisters:
> But when I was looking for this information in the documentation, I never saw the sentence you refer to. It is apparently hidden by an indirection.

One idea would be to explicitly mention import in the existing sentence, since that might have helped direct you to the correct page. Maybe changing 

> Both user-defined and internal presets can be viewed and managed from within [preferences > presets](../../../preferences-settings/presets.md).

to

> Both user-defined and internal presets can be viewed, managed, and imported from within [preferences > presets](../../../preferences-settings/presets.md).

Issue #947, opened 2026-07-01
Document highlight protection mode in denoise(profiled)


anoderay:
This needs documentation:

In denoise(profiled), the autodetected noise profile is adjusted to a lower ISO when enabled, reflecting the fact that the image was shot at a lower ISO than recorded in the Exif data.  As in the exposure module, there is a checkbox to disable the compensation, which is hidden when there is none to be applied.

_Originally posted by @ralfbrown in https://github.com/darktable-org/darktable/issues/19375#issuecomment-4855758489_

anoderay:
Closed by #1021

PR #1021, opened 2026-07-29
Add note to denoise about hidden exposure compensation

Phemisters:
Relates to dtdocs issue: https://github.com/darktable-org/dtdocs/issues/947

Add to _denoise (profiled)_ about the highlight compensation checkbox (only visible if camera adjusted exposure in certain modes).

Also re-arranged order of some options to match what I saw during testing.


anoderay:
Looks good, thanks!

PR #1022, opened 2026-07-29
Describe shape rotation

Phemisters:
Relates to darktable issue: https://github.com/darktable-org/darktable/pull/21349

Mention shape rotation is now generally available for mask shapes (was previously just ellipse).


anoderay:
I can't test currently: are you 100% it is just within a shape? Can't test currently :-). If yes: merge away!

Phemisters:
> are you 100% it is just within a shape

Certainly works that way in my tests - if I start the drag outside the shape it does not trigger a rotation. Once the rotation starts, the mouse can move outside the shape, but that feels natural enough not to need documented. I'll merge it as it is.

PR #1034, opened 2026-08-04
Make translations sort on name

Phemisters:
When I added Lao to the languages I simply added it to the end of the list, and incremented the weight given to Dutch. I see now that Dutch and Lao appear at the end of the list which seems odd. Instead of having to always change a lot of weights when any new language is added, this PR attempts to sort the languages by name, which seems sensible to me.


anoderay:
Good idea, thanks!

PR #1036, opened 2026-08-04
Describe image choice for darkroom

Phemisters:
Relates to darktable issue: https://github.com/darktable-org/darktable/pull/21613

I didn't find any existing description of which image gets used, so I tried to add something about this. Open to suggestions of a better place or description.

### DIFF
--- a/content/darkroom/overview.md
+++ b/content/darkroom/overview.md
@@ -4,7 +4,17 @@ id: overview
 weight: 10
 ---
 
-The darkroom view is where you develop your images. The center panel contains the image currently being edited.
+The darkroom view is where you develop your images. There are multiple ways to start developing an image in the darkroom:
+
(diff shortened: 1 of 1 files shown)


anoderay:
Looks good, thanks!

anoderay:
@Phemisters do you get why the gh pages deployment failed? The check was fine.

Phemisters:
> do you get why the gh pages deployment failed? The check was fine.

@anoderay  I'm not sure. My only idea at the moment is that I see two PRs were merged within seconds of each other, and I wonder if there were two deploy actions running at the same time and somehow conflict with each other.

anoderay:
Maybe that was it. Anyway the last deployment following merge of translations went through so there seems to not be a systemic issue.

PR #1038, opened 2026-08-05
Add doc for OpenCL nocl option

Phemisters:
Relates to darktable PR: https://github.com/darktable-org/darktable/pull/21287

The darktable PR says:

> 1. remove the advantage feature
> 2. introduce a per-device entry in the conf database 'device_nocl This can hold a comma-separated list of module->so names. Any module in this list will be executed on the CPU instead of OpenCL.

This doc PR tries to describe the changes.

It's annoying to be adding notes about _when_ a change becomes relevant, but until we have doc versions being published I'm not sure what else to do.


anoderay:
Yeah I agree on the versioning issue... Thanks anyway :-)!

PR #1056, opened 2026-08-14
Remove mention of preferences for slider-color

anoderay:
The option has apparently been removed in the preferences - at least I can't find it.


victoryforce:
You are right, it was removed: https://github.com/darktable-org/darktable/pull/12112

I just noticed that the PR title for some reason contained an error, but the PR itself removed the `colored_sliders` key from the UI, not the `button_bar`.

PR #1059, opened 2026-08-16
feat: clarify using test environment

yuri1969:
This is an attempt to improve clarity and ease of usage of the *using a test environment* section.

* Made usage of in-memory database clear
* Provided launch option protecting XMP files
* Provided TL;DR command line options
* Added links, etc.

### DIFF
--- a/content/special-topics/test-environment.md
+++ b/content/special-topics/test-environment.md
@@ -77,15 +77,35 @@ b) Turn off XMP creation in the other installation's preferences
 
 For an independent installation that should not write XMP sidecar files:
 
-```
-darktable --configdir /path-to-test-config-dir --conf write_sidecar_files=never
-```
+Linux/macOS
+: Options:
+
+  ```
+  --configdir /path-to-test-config-dir --conf write_sidecar_files=never
+  ```
+
+Windows
+: Options:
+
+  ```
+  --configdir "C:\path-to-test-config" --conf "write_sidecar_files=never"
+  ```
 
 To additionally keep all changes to its library database in memory only:
 
-```
-darktable --configdir /path-to-test-config-dir --conf write_sidecar_files=never --library :memory:
-```
+Linux/macOS
+: Options:
+
+  ```
+  --configdir /path-to-test-config-dir --conf write_sidecar_files=never --library :memory:
+  ```
+
+Windows
+: Options:
+
+  ```
+  --configdir "C:\path-to-test-config" --conf "write_sidecar_files=never" --library ":memory:"
+  ```
 
 For the safest setup, use copies of the RAW files rather than relying on XMP creation being disabled.
 
--- a/content/special-topics/test-environment.md
+++ b/content/special-topics/test-environment.md
@@ -29,7 +29,7 @@ darktable stores its settings, presets, styles, and library database in its [con
 
 To prevent an installation from writing to the same configuration directory as another installation, launch it with the option `--configdir`, followed by the path to a new folder. If the folder does not exist yet, darktable creates and populates it automatically.
 
-You can run an experiment without persistent library changes by additionally setting `--library :memory:`, which keeps the library database in RAM. All changes to the database are discarded on exit: `darktable --configdir /path/to/test-config --library :memory:`
+You can run an experiment without making persistent changes to the library by specifying `--library :memory:`, which keeps the library database in RAM. All changes to the database are discarded on exit: `darktable --configdir /path/to/test-config --library :memory:`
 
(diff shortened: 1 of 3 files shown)


content/special-topics/test-environment.md
```diff
@@ -58,17 +63,30 @@ macOS
...
+
+b) Turn off XMP creation in the other installation's preferences
+: Launch darktable with the additional `--conf write_sidecar_files=never` option. Alternatively, you can set _preferences / storage / create XMP files_ to "never" in the UI. This lets you open folders with existing edits without writing changes back to their XMP files. The setting only affects this installation because it has its own configuration.
+
+## Summary
+
+For an independent installation that should not write XMP sidecar files:
+
+```
+darktable --configdir /path-to-test-config-dir --conf write_sidecar_files=never
```
anoderay: This is only accurate for Linux package installations and could be confusing for windows users.

anoderay (changes requested):
Thanks for your contribution! You improved the article in many areas! One comment for your consideration.

content/special-topics/test-environment.md
yuri1969: The options examples of the _Summary_ section got split to Windows & Linux/MacOS.

content/special-topics/test-environment.md
```diff
@@ -58,17 +63,50 @@ macOS
...
+
+  ```
+  --configdir "C:\path-to-test-config" --conf "write_sidecar_files=never" --library ":memory:"
+  ```
 
-For a lasting test setup, options a) or b) are practical. Option c) is handy for a quick, no-trace experiment.
+For the safest setup, use copies of the RAW files rather than relying on XMP creation being disabled.
 
-**Caution:** If you cloned your main installation's configuration directory for the testing environment, make sure to either disable XMP creation as described above under b) or remove (not delete) all photos from the library in your testing environment. Otherwise you could interfere with your main installation's edits.
+**Caution:** If you cloned an existing installation's configuration directory for another installation, disable XMP file creation as described above under b). Alternatively, remove (not delete) all photos from the library in the new installation. This removes them only from the library database; it does not delete the files from disk. Otherwise, you could interfere with edits made by the original installation.
```
victoryforce: This sounds like over-explanation. When we suggest "remove all photos from the library", it means exactly what it says. There is no point in clarifying that this does not mean deleting files from the disk.

content/special-topics/test-environment.md
```diff
@@ -58,17 +63,50 @@ macOS
...
+
+Windows
+: Options:
+
+  ```
+  --configdir "C:\path-to-test-config" --conf "write_sidecar_files=never" --library ":memory:"
+  ```
 
-For a lasting test setup, options a) or b) are practical. Option c) is handy for a quick, no-trace experiment.
+For the safest setup, use copies of the RAW files rather than relying on XMP creation being disabled.
```
victoryforce: These may not be RAW files, there is nothing that prevents you from processing JPEG, HEIC, or any other non-RAW formats. Therefore, instead of "RAW files" it should be "image files", and not only here but also in several cases above (which were in the text before this PR).

content/special-topics/test-environment.md
```diff
@@ -4,41 +4,46 @@ id: test-environment
...
 
-To stop the second installation from writing into that same directory, launch it with the option `--configdir`, followed by the path to a new folder. If it doesn't exist yet, darktable creates and populates it automatically.
+To prevent an installation from writing to the same configuration directory as another installation, launch it with the option `--configdir`, followed by the path to a new folder. If the folder does not exist yet, darktable creates and populates it automatically.
 
-Tip: You can also copy your existing configuration directory into the new folder instead of starting from scratch. That way the test version inherits your presets, styles, and keyboard shortcuts as a starting point, while both installations still run fully independently from there on. This requires special care in step 3 to avoid affecting your existing edits.
+You can run an experiment without persistent library changes by additionally setting `--library :memory:`, which keeps the library database in RAM. All changes to the database are discarded on exit: `darktable --configdir /path/to/test-config --library :memory:`
```
Phemisters: To me a setting is something in preferences which can be set once.  This might be better as _...additionally specifying..._.

yuri1969: The sentence got rephrased

content/special-topics/test-environment.md
anoderay: I'd make it `Step 3b` instead of `b)` because with your additions there is a bit of distance between this "caution"-note and the references step.

anoderay (changes requested):
Otherwise fine to merge from my side!

Phemisters:
Thank you - merged.

PR #1060, opened 2026-08-16
refined documentation for contrast & texture: edge protection

anoderay:
Based on [this answer](https://discuss.pixls.us/t/merged-new-module-contrast-and-texture/59768/63?u=mino) by @jandren I tried to refine the documentation for edge protection.

Side-note @jandren : I think this setting should be called "edge detection" because "protection" feels off in the contrast of managing local contrast.


Phemisters:
Thanks - looks good to me.

PR #1080, opened 2026-08-29
Add documentation on finding basic adjustments

stnKrisna:
Add a new section to the processing workflow guide for users who are already familiar with common image-editing terminology. The wording is intentionally application-neutral so the guide is useful to users transitioning from any editor that uses similar terminology.

My changes aims to lowers the learning curve without changing darktable’s processing model or encouraging users to treat it like another application.
Its main value is that it bridges familiar editing terminology with darktable’s actual tools and scene-referred workflow.

- explaining why darktable does not always use one-to-one slider equivalents and a look-up table for the basic adjustments (adjust highlights, shadows, clarity, vibrance, dehaze) and the darktable module
- pointing users toward the small set of modules that cover most routine edits
- reducing trial-and-error and unnecessary module searching
- keeping the documentation application-neutral, so it benefits users coming from many different editors

### DIFF
--- a/content/overview/workflow/process.md
+++ b/content/overview/workflow/process.md
@@ -82,6 +82,75 @@ For the purposes of this guide, click on the hamburger icon (to the right of the
 
 ---
 
+## finding familiar adjustments
+
+If you have previously used other image-processing software, you may already be familiar with controls named *exposure*, *contrast*, *highlights*, *shadows*, *whites*, *blacks*, *texture*, *clarity*, *dehaze*, *vibrance* and *saturation*.
+
+darktable does not always provide a one-to-one equivalent for these controls. In particular, scene-referred processing separates adjustments to the brightness of the scene from the later mapping of that scene onto the display. As a result, an adjustment that is represented by a single slider elsewhere may be performed by a different module depending on what you are trying to achieve.
+
+The following table provides useful starting points.
+
+| familiar adjustment | darktable module/control                                                                                                                                                                         | notes                                                                                                                                                                                                                                                                     |
+| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| **exposure**        | [exposure](../../module-reference/processing-modules/exposure.md) → **exposure**                                                                                                                 | Adjust the overall brightness of the image. This should normally be used to establish the desired mid-tone brightness early in the edit.                                                                                                                                  |
+| **contrast**        | [tone equalizer](../../module-reference/processing-modules/tone-equalizer.md), or the contrast control of the active display transform                                                           | Use tone equalizer when you want to redistribute scene brightness. If using AgX, its **contrast** control adjusts the slope of the display tone curve.                                                                                                                    |
+| **highlights**      | [tone equalizer](../../module-reference/processing-modules/tone-equalizer.md) → brighter exposure zones                                                                                          | Lower the brighter zones to recover or compress bright areas, or raise them to make highlights more prominent.                                                                                                                                                            |
+| **shadows**         | [tone equalizer](../../module-reference/processing-modules/tone-equalizer.md) → darker exposure zones                                                                                            | Raise the darker zones to open shadows or lower them to make shadows deeper.                                                                                                                                                                                              |
+| **whites**          | display transform white point; for AgX use **white relative exposure**                                                                                                                           | This establishes which scene values are mapped towards display white. If you merely want to brighten the brightest objects without altering the white point, use the brighter zones of tone equalizer instead.                                                            |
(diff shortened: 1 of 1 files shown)


content/overview/workflow/process.md
```diff
@@ -82,6 +82,75 @@ For the purposes of this guide, click on the hamburger icon (to the right of the
...
+| **exposure**        | [exposure](../../module-reference/processing-modules/exposure.md) → **exposure**                                                                                                                 | Adjust the overall brightness of the image. This should normally be used to establish the desired mid-tone brightness early in the edit.                                                                                                                                  |
+| **contrast**        | [tone equalizer](../../module-reference/processing-modules/tone-equalizer.md), or the contrast control of the active display transform                                                           | Use tone equalizer when you want to redistribute scene brightness. If using AgX, its **contrast** control adjusts the slope of the display tone curve.                                                                                                                    |
```
anoderay, line 96: I think sending newcomers to TE for contrast adjustment could be overwhelming. "How to add contrast" is a simple question with lots of complex answers in dt and I think deserves a longer answer: E.g. "In Sigmoid use X, or "contrast" in Color Balance RGB (but note that it primarily affects brightness and not colours).

content/overview/workflow/process.md
```diff
@@ -82,6 +82,75 @@ For the purposes of this guide, click on the hamburger icon (to the right of the
...
+| **highlights**      | [tone equalizer](../../module-reference/processing-modules/tone-equalizer.md) → brighter exposure zones                                                                                          | Lower the brighter zones to recover or compress bright areas, or raise them to make highlights more prominent.                                                                                                                                                            |
```
anoderay, line 97: Shadows/Highlights deserves a mention, too. I think. Maybe with the caveat that it is prone to artifacting when doing big adjustments.

content/overview/workflow/process.md
```diff
@@ -82,6 +82,75 @@ For the purposes of this guide, click on the hamburger icon (to the right of the
...
+| **shadows**         | [tone equalizer](../../module-reference/processing-modules/tone-equalizer.md) → darker exposure zones                                                                                            | Raise the darker zones to open shadows or lower them to make shadows deeper.                                                                                                                                                                                              |
+| **whites**          | display transform white point; for AgX use **white relative exposure**                                                                                                                           | This establishes which scene values are mapped towards display white. If you merely want to brighten the brightest objects without altering the white point, use the brighter zones of tone equalizer instead.                                                            |
```
anoderay, line 99: what do you mean by "display transform white point"?

anoderay (commented):
Hey @stnKrisna ! Thanks for taking the time improving this article! 

The article further down sends newcomers to "filmic" though the current default is Sigmoid. That part (image processing in 3 modules) could use a do-over as well considering sigmoid is the default for new users. 

I would suggest tweaking your table assuming sigmoid as the set tonemapper. AgX could IMHO overwhelm new users with it's many controls.

Phemisters:
I don't  have enough experience of other apps to say anything much about the content. I do, however,  think the _process_ page is already quite long, so perhaps make this a separate  page. That would also make it more easily discoverable by new users.

anoderay:
You are right, that article is fairly long. Maybe split up into: 
* Getting started
* why is the jpeg different from the raw?
* Processing in 3 modules
* Other processing techniques

I'd put "other important topics" under "getting started".

mrjohnc:
@stnKrisna @Phemisters @anoderay anoderay etc I would be happy to work on you with this. I am fairly new to Darktable but have much more experience with Lightroom. Maybe we can put this in a place that is easy for people to contribute to and then we find a better home once it is more progressed? Personally I think a comparison to other editors like Lightroom would be important to have on the website to support people in migrating from Lightroom etc.

Thanks :)

Phemisters:
> Maybe we can put this in a place that is easy for people to contribute to and then we find a better home once it is more progressed?

I don't think it needs a separate place - it is easy for people to raise subsequent dtdocs PRs once we have something in place. And the web site can simply link to the manual page, if we need to increase the visibility.

Or do you mean you want to contribute to this PR? For now, you can maybe just do that with specific comments on the currently suggested content.

mrjohnc:
> > Maybe we can put this in a place that is easy for people to contribute to and then we find a better home once it is more progressed?
> 
> I don't think it needs a separate place - it is easy for people to raise subsequent dtdocs PRs once we have something in place. And the web site can simply link to the manual page, if we need to increase the visibility.
> 
> Or do you mean you want to contribute to this PR? For now, you can maybe just do that with specific comments on the currently suggested content.

Whatever works best for others, can someone link to where the current draft is? Also is there a list of Darktable features already created which could form the Darktable part of the table?

anoderay:
@mrjohnc the current draft for the propsed changes is this PR. Under "[files changes](github.com/darktable-org/dtdocs/pull/1080/changes)" you can see the current state. 

Word of caution: I don't know how far you've wandered into darktable land yet. These comparison tables "lr vs dt" have a fundamental problem: dt is *fundamentally* different from LR and requires a fair bit of relearning.

PR #1097, opened 2026-09-04
[5.6] fix version switcher

anoderay:
was broken, this fixes it

Co Created with Claude Sonnet 5

### DIFF
--- a/themes/hugo-darktable-docs-theme/layouts/partials/page-header-navbar-large.html
+++ b/themes/hugo-darktable-docs-theme/layouts/partials/page-header-navbar-large.html
@@ -18,7 +18,7 @@
     <ul class="navbar-nav ml-auto">
       {{ if .IsTranslated }}
         <li class="nav-item dropdown">
-          <a
+
             class="nav-link dropdown-toggle"
             href="#"
             data-toggle="dropdown"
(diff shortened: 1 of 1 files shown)


anoderay:
Hm, the fix did work locally but not on the deployment. Will investigate.

Phemisters:
> Hm, the fix did work locally but not on the deployment. Will investigate.

@anoderay  I was looking a the same problem when I noticed you were already working on it. Again, I have a solution which works locally but no idea f it will work when deployed. I'll put it here for consideration, but you may find a better solution:

```html
<div class="dropdown-menu dropdonw-trans" role="listbox">
            {{ $cleanPath := strings.ReplaceRE "/usermanual/[^/]*/" "/" .RelPermalink  }}
            <a class="dropdown-item" href="/usermanual/development{{ $cleanPath }}">Development (Master)</a>
            <a class="dropdown-item" href="/usermanual/5.6{{ $cleanPath }}">Version 5.6</a>
            <a class="dropdown-item" href="/usermanual/4.8{{ $cleanPath }}">Version 4.8</a>
            <a class="dropdown-item" href="/usermanual/4.6{{ $cleanPath }}">Version 4.6</a>
            <a class="dropdown-item" href="/usermanual/4.4{{ $cleanPath }}">Version 4.4</a>
</div>

```

anoderay:
Found it, somehow managed to delete an `<a` - fixed in https://github.com/darktable-org/dtdocs/commit/7560a1bd4de2e2df05fece77b48a183eaefe6e0f

PR #1098, opened 2026-09-04
[master] fix version switcher

anoderay:
version switcher fix as in https://github.com/darktable-org/dtdocs/pull/1097


Phemisters:
Looks good, merge if you are ready.

PR #1102, opened 2026-09-05
Document the cache dir

lefth:
Currently, the path in the docs is ~/.cache/darktable. This isn't applicable to Windows or all Linux setups.

I didn't include a path for snap installations. Does someone know? I'll be happy to add it.


content/special-topics/program-invocation/darktable.md
```diff
@@ -75,10 +75,10 @@ All parameters are optional. In most cases darktable should be started without a
 : Define where darktable should store its temporary files. If this option is not supplied darktable uses the system default.
 
 `--cachedir <cache directory>`
-: darktable keeps a cache of image thumbnails for fast image preview and precompiled OpenCL binaries for fast startup. By default the cache is located in `$HOME/.cache/darktable/`. Multiple thumbnail caches may exist in parallel -- one for each library file.
+: darktable keeps a cache of image thumbnails for fast image preview and precompiled OpenCL binaries for fast startup. Multiple thumbnail caches may exist in parallel -- one for each library file. The default cache path is listed on the [configuration directory](../../preferences-settings/config-directory.md) page.
```
Phemisters: Perhaps make this link go directly to the cache part of the page, just like the links in other pages.

content/preferences-settings/config-directory.md
```diff
@@ -8,17 +8,17 @@ darktable stores its settings, presets, styles, library database, etc. in a defa
...
+## Default configuration paths
 
-* Linux 
+* Linux
     + Installed via package manager or AppImage: `$HOME/.config/darktable`
     + Flatpak: `$HOME/.var/app/org.darktable.Darktable/config/darktable`
     + snap: `$HOME/snap/darktable/current/`
 * Windows: `%LOCALAPPDATA%\darktable`
-	+ E.g. `C:\Users\<username>\AppData\Local\darktable`
+	+ I.e., `C:\Users\<username>\AppData\Local\darktable`
```
Phemisters: It has been a while since I used Windows, but is not still possible to move the user area to another drive? In which case, _E.g._ would be correct., since it might be that _D:_ (for example) would be correct. And now that you have drawn attention to it, it might read better as _Typically_ or something.

Phemisters (commented):
Looks fine to me - I've left a couple of comments for consideration.

lefth:
Thanks, @Phemisters. I pushed an update with your suggestions.

Phemisters:
Thank you - looks good to me.

PR #1120, opened 2026-09-14
move styles to shared utility modules to "shared"

anoderay:
because it is a shared module.


Phemisters:
@anoderay  At line 31 we have

> mode
: As with the [history stack](./history-stack.md) module, this combobox allows you to either "append" the style to the current history stack or to "overwrite" the history stack of the target image.

and I think the history stack link here needs to be the _lighttable_ one: `[lighttable history stack](../lighttable/history-stack.md)`

Once that is changed I think this is ready for merging.

anoderay:
Thanks for the review @Phemisters .

PR #1123, opened 2026-09-15
add caution for unsupported metadata settings for some export formats

anoderay:
Adresses the issue discussed in https://github.com/darktable-org/darktable/issues/22181 . @victoryforce


content/module-reference/utility-modules/shared/export.md
```diff
@@ -36,7 +36,7 @@ on conflict
 ## format options
 
 file format
-: Choose the file format for the exported image. Additional options will appear (below) depending on the selected format.
+: Choose the file format for the exported image. Additional options will appear (below) depending on the selected format. Note that for some formats there are caveats concerning metadata inclusion, see [below](#metadata-preferences).
```
Phemisters: In general, best practice is to make the link text meaningful (that is, not things like _here_ or _below_). I don't know if we try to enforce that in the docs.

https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html

Phemisters (commented):
I've left one comment for consideration, but I'm happy for this to be merged.

content/module-reference/utility-modules/shared/export.md
anoderay: I agree, "below" is not very elegant (better then click this [link](link.html) though :-D) - updated the branch.

PR #1133, opened 2026-09-20
defuse wording concerning combinations of tonemappers

anoderay:
In the discussion about my proposal of [warning against multiple tonemappers](https://github.com/darktable-org/darktable/issues/22237) multiple maintainers voiced that combining tonemappers is a valid usecase (albeit not the primarily intended). Hence I would like to defuse the wording in the tonemappers articles. 

"Never" to me implies "bad things happen if you do" - but that truly is not the case.


Phemisters:
A lot of the discussion in that darktable PR was new to me. I wonder if, instead of each tone mapper page having a list of other tone mappers (which will all have to be updated if another tone mapper gets added), we should perhaps add a brief page about tone mappers. It would, I think, be short - just list all the existing tone mappers, have a sentence about what makes a tone mapper (whatever that is!), and a note that most users only need one tone mapper active at a time.

Then, the individual tone mappers could all have a consistent sentence along the lines of: _This is a tone mapper, you generally need only one active at a time_, along with a link to the tone mapper page.

What do you think?

anoderay:
Good idea. Within that there is the question whether we call these display transforms or tonemappers. I'd strongly prefer the latter but historically it's been "display transform".

@TurboGit could you chime in on that question?

TurboGit:
> A lot of the discussion in that darktable PR was new to me. I wonder if, instead of each tone mapper page having a list of other tone mappers (which will all have to be updated if another tone mapper gets added), we should perhaps add a brief page about tone mappers. It would, I think, be short - just list all the existing tone mappers, have a sentence about what makes a tone mapper (whatever that is!), and a note that most users only need one tone mapper active at a time.

Yes, good idea to me too.

> I'd strongly prefer the latter but historically it's been "display transform".

I'm ok with "Tone mapper" but anyway you'll probably need to explain that a tone mapper is a way to transform the - scene referred picture - for the display. This is a very important point for people to understand, until the tone mapper is applied the image RVB values are unbounded and just invalid for the display.

Keep up the good work on the documentation.

anoderay:
@Phemisters are you fine with merging this?

Phemisters:
> @Phemisters are you fine with merging this?

@anoderay if we plan on adding a tone mapper page and making these pages all refer to it, is it worth making this change? If you think it is, then sure, merge it as it currently is.

anoderay:
Considering this change is two clicks away I will merge it. Do you have the "tonemapper page" on your todo-list or should I?

Phemisters:
> Considering this change is two clicks away I will merge it. Do you have the "tonemapper page" on your todo-list or should I?

I'll take a look in a few days.

